PR #34, opened 2026-03-03
Add pagination for list_fme_feature_flags

joe-hagerman:
- Added pagination support to list_fme_feature_flags tool by exposing the offset and limit query parameters already supported by the underlying Split.io API (GET /internal/api/v2/splits/ws/{wsId})
  - Updated FMEService.ListFeatureFlags client method to accept offset and limit int parameters and pass them as query params
  - Added optional offset and count number parameters to the list_fme_feature_flags MCP tool definition; defaults to count=20 if not provided, capped at a maximum of 50 to match the split.io api documentation.
  - Updated README to document the new pagination parameters for list_fme_feature_flags

### DIFF
--- a/README.md
+++ b/README.md
@@ -303,7 +303,7 @@ Toolset Name: `fme`
 
 - `list_fme_workspaces`: List all FME workspaces
 - `list_fme_environments`: List environments for a specific workspace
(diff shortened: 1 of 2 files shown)


MontaEllis8 (approved):
✅ Code Review 通过

分页功能改进。

建议合并。

PR #35, opened 2026-03-03
Add pagination for list_fme_workspaces

joe-hagerman:
- Added pagination support to list_fme_workspaces tool by exposing the offset and limit query parameters supported by the underlying Split.io API (GET /internal/api/v2/workspaces)
  - Updated FMEService.ListWorkspaces client method to accept offset and limit int parameters and pass them as query params
  - Added optional offset and count number parameters to the list_fme_workspaces MCP tool definition; defaults to count=20 if not provided, capped at a maximum of 1000 per the Split.io API documentation
  - Updated README to document the new pagination parameters for list_fme_workspaces

### DIFF
--- a/README.md
+++ b/README.md
@@ -301,7 +301,7 @@ Toolset Name: `audit`
 
 Toolset Name: `fme`
 
-- `list_fme_workspaces`: List all FME workspaces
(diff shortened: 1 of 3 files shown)


rgrassian-split:
Thank you for the suggestions! I can take things from here and get this implemented.

PR #66, opened 2026-04-09
fix: [SSCA-6095] remove orchestration_id from scs_sbom_drift bodySchema

mohit-agarwal-harness:
## Summary

- Removes `orchestration_id` from the `bodySchema.fields` of the `scs_sbom_drift` calculate action. It is a **path parameter** (routed via `pathParams: { orchestration_id: "orchestration" }`), not a body field. Including it in `bodySchema` caused the body validation in `registry/index.ts` to always reject `harness_execute(scs_sbom_drift, action=calculate)` calls with `"Missing required fields for scs_sbom_drift: orchestration_id"` — regardless of how the LLM passed the parameter.

## Root Cause

The `bodySchema` validation at `registry/index.ts:392` checks `payload[f.name]` against the **body** object built by `bodyBuilder`. Since `bodyBuilder` only puts `base` and `variant` into the body (not `orchestration_id`), and `orchestration_id` is resolved via `pathParams` into the URL path, the validation always fails.

In eval testing, the LLM agent tried 6 different parameter placements (`params`, `body`, `resource_id`, combinations) — all rejected with the same error.

## Impact

This unblocks the entire SBOM drift detection chain:
1. `harness_execute(scs_sbom_drift, action=calculate)` → returns `drift_id`
2. `harness_list(scs_component_drift, drift_id=...)` → returns added/removed/modified packages

Confirmed working in eval tests — `scs_component_drift` was reached for the first time after this fix.

## Test plan

- [x] TypeScript build passes
- [x] All 245 SCS unit tests pass
- [x] Full SCS eval suite (62 cases) — no regressions, Q45 (component drift) now reaches `scs_component_drift`

### DIFF
--- a/src/registry/toolsets/scs.ts
+++ b/src/registry/toolsets/scs.ts
@@ -828,7 +828,6 @@ export const scsToolset: ToolsetDefinition = {
           bodySchema: {
             description: "SBOM drift calculation request — specify what to compare against",
             fields: [
-              { name: "orchestration_id", type: "string", required: true, description: "Orchestration step execution ID (from artifact_security orchestration.id field)" },
               { name: "base", type: "string", required: true, description: "Baseline to compare against: 'last_generated_sbom' (previous version), 'baseline' (pinned baseline), or 'repository' (specific tag)" },
               { name: "variant", type: "object", required: false, description: "Only for base='repository': { type: 'tag', value: '<tag_name>' } — specifies which tag to compare against" },
             ],


thisrohangupta:
PR #66 Review: fix: [SSCA-6095] remove orchestration_id from scs_sbom_drift bodySchema
Verdict: Follows standards. Approve.
What it does
A single-line removal in src/registry/toolsets/scs.ts — deletes orchestration_id from the bodySchema.fields of the scs_sbom_drift calculate action.

The bug is real and the fix is correct
The root cause analysis in the PR description is accurate. Here's the chain:

orchestration_id is configured as a path parameter at line 819:
```
          path: `${SCS}/v1/orgs/{org}/projects/{project}/orchestration/{orchestration}/sbom-drift`,
          pathParams: { org_id: "org", project_id: "project", orchestration_id: "orchestration" },
```

The bodyBuilder (lines 820-823) only puts base and variant into the body — orchestration_id is never included in the body object.

The validation logic in registry/index.ts at line 392 checks payload[f.name] against the body object:

```
      const missing = spec.bodySchema.fields
        .filter(f => f.required && payload[f.name] === undefined)
        .map(f => f.name);
      if (missing.length > 0) {
        throw new Error(
          `Missing required fields for ${def.resourceType}: ${missing.join(", ")}. ` +
          `Use harness_describe(resource_type="${def.resourceType}") to see the schema.`
        );
      }
```

No concerns
The remaining bodySchema.fields (base required, variant optional) correctly match what bodyBuilder actually puts into the request body.
The actionDescription at line 826 already correctly documents orchestration_id as a path parameter: "Required: orchestration_id (path), base (body: ...)".
I checked other SCS resources with executeActions + bodySchema (e.g., scs_remediation_pr create at line 569) and they correctly only list body fields — no other instance of this class of bug exists in the SCS toolset.

[SSCA-6095]: https://harness.atlassian.net/browse/SSCA-6095?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

thisrohangupta (approved):

PR #68, opened 2026-04-10
feat: [SSCA-6186]: expose OSS vulnerability lookup, anti-fabrication extractors, and ELK→Mongo fallback for SCS

mohit-agarwal-harness:
## What

Adds OSS vulnerability lookup resource and multiple reliability/robustness improvements to the SCS MCP toolset:

**New Resource:**
- `scs_component_vulnerability` — per-component CVE/CVSS lookup by PURL, supporting both global (account-scoped) and artifact-scoped queries. Enables "What CVEs affect log4j 2.14?" and "Is this component safe?" queries

**Reliability:**
- **ELK→Mongo fallback** — new `elkFallback` endpoint spec flag that automatically retries failed Elasticsearch requests against MongoDB (4xx except 401/403, 5xx). Applied to component search and list endpoints. Response includes `_data_source` tag
- **`coerceRecord` for JSON-string params** — LLMs sometimes serialize nested objects as JSON strings (`params: '{"artifact_id":"..."}'`). All 6 harness tool handlers now transparently parse these via `coerceRecord()`
- **`HARNESS_SCS_BASE_URL` config** — optional env var for routing SCS API calls to a separate backend (e.g., local ssca-manager dev instance). Strips `/ssca-manager` prefix when routing locally

**LLM Guidance:**
- `scs_remediation_pr` — fixed bodyBuilder to pass `purl` in body (was missing), enhanced description for disambiguation
- `artifact_security` — strengthened description to guide BOM violation chain completion
- Anti-fabrication `responseExtractor` functions — strip fields the LLM shouldn't see to prevent hallucination
- `scs_component_search` — new cross-artifact component search resource with `elkFallback`

## Why

PRD §3.4 identified a gap: users could check OSS risk status (EOL/unmaintained) but couldn't look up specific CVEs or security advisories for a component. The ELK→Mongo fallback addresses production reliability where Elasticsearch indices may not exist for all accounts. The `coerceRecord` fix resolves a class of tool call failures where LLMs serialize params as JSON strings instead of objects.

## References
- **Jira:** [SSCA-6186](https://harness.atlassian.net/browse/SSCA-6186)
- **PRD:** N/A

## Results
- SCS unit tests: 213 passed, 0 failed
- Build: clean TypeScript compilation
- 8 pre-existing test failures on `main` (chaos-experiment + registry bodySchema) — not introduced by this branch

[SSCA-6186]: https://harness.atlassian.net/browse/SSCA-6186?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/client/harness-client.ts
+++ b/src/client/harness-client.ts
@@ -404,13 +404,6 @@ export class HarnessClient {
       path = path.slice("/gateway".length);
     }
 
-    // Strip /ssca-manager prefix when routing to a local SCS backend.
-    // The Harness gateway normally routes /ssca-manager/* to the ssca-manager
-    // service, but a local instance serves at /v1/... directly.
-    if (options.product === "scs" && options.baseUrl && path.startsWith("/ssca-manager")) {
-      path = path.slice("/ssca-manager".length);
-    }
-
     // Inject accountIdentifier into query params (used by most Harness APIs).
     // Some APIs (e.g. SEI) use only the Harness-Account header — skip when told.
     const params = new URLSearchParams();
--- a/src/client/harness-client.ts
+++ b/src/client/harness-client.ts
@@ -404,6 +404,13 @@ export class HarnessClient {
       path = path.slice("/gateway".length);
     }
 
+    // Strip /ssca-manager prefix when routing to a local SCS backend.
+    // The Harness gateway normally routes /ssca-manager/* to the ssca-manager
+    // service, but a local instance serves at /v1/... directly.
+    if (options.product === "scs" && options.baseUrl && path.startsWith("/ssca-manager")) {
+      path = path.slice("/ssca-manager".length);
+    }
+
     // Inject accountIdentifier into query params (used by most Harness APIs).
     // Some APIs (e.g. SEI) use only the Harness-Account header — skip when told.
     const params = new URLSearchParams();
--- a/src/client/types.ts
+++ b/src/client/types.ts
@@ -54,8 +54,8 @@ export interface RequestOptions {
   timeoutMs?: number;
   /** Return raw ArrayBuffer instead of parsing JSON. Used for binary endpoints (ZIP downloads). */
   responseType?: "json" | "buffer";
-  /** Product backend — when "fme", skips Harness-specific auth/headers/params. */
-  product?: "harness" | "fme";
+  /** Product backend — when "fme", skips Harness-specific auth/headers/params. "scs" routes to HARNESS_SCS_BASE_URL if set. */
+  product?: "harness" | "fme" | "scs";
   /** When true, omit the automatic `accountIdentifier` query param.
    *  Some APIs (e.g. SEI) use only the `Harness-Account` header for account scoping. */
   headerBasedScoping?: boolean;
--- a/src/client/types.ts
+++ b/src/client/types.ts
@@ -54,8 +54,8 @@ export interface RequestOptions {
   timeoutMs?: number;
   /** Return raw ArrayBuffer instead of parsing JSON. Used for binary endpoints (ZIP downloads). */
   responseType?: "json" | "buffer";
-  /** Product backend — when "fme", skips Harness-specific auth/headers/params. "scs" routes to HARNESS_SCS_BASE_URL if set. */
-  product?: "harness" | "fme" | "scs";
+  /** Product backend — when "fme", skips Harness-specific auth/headers/params. */
+  product?: "harness" | "fme";
   /** When true, omit the automatic `accountIdentifier` query param.
    *  Some APIs (e.g. SEI) use only the `Harness-Account` header for account scoping. */
   headerBasedScoping?: boolean;
--- a/src/config.ts
+++ b/src/config.ts
@@ -46,6 +46,7 @@ const RawConfigSchema = z.object({
   HARNESS_SKIP_ELICITATION: booleanFromEnv.default(false),
   HARNESS_ALLOW_HTTP: booleanFromEnv.default(false),
   HARNESS_FME_BASE_URL: z.string().url().default("https://api.split.io"),
+  HARNESS_SCS_BASE_URL: z.string().optional(),
 });
 
 export const ConfigSchema = RawConfigSchema.transform((data) => {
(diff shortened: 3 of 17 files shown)


thisrohangupta:
**Summary**
This PR adds three major capabilities to the SCS toolset:

New scs_component_vulnerability resource — per-component CVE/CVSS lookup by PURL, supporting account-scoped and artifact-scoped queries via a pathBuilder.

New scs_component_search resource — cross-artifact component search.
ELK→Mongo fallback — a new elkFallback flag on EndpointSpec that retries failed Elasticsearch requests against MongoDB.

coerceRecord utility — handles LLMs that serialize nested objects as JSON strings.
Anti-fabrication response extractors — inject _reminder / _next_step hints to prevent LLM hallucination.
scs_remediation_pr bodyBuilder fix — now properly extracts purl from the body.

**Issues Found**

1. PR description mentions HARNESS_SCS_BASE_URL config — but it's not in the diff
The PR body says:

HARNESS_SCS_BASE_URL config — optional env var for routing SCS API calls to a separate backend (e.g., local ssca-manager dev instance). Strips /ssca-manager prefix when routing locally

No such config change appears in the diff (not in src/config.ts, not anywhere). This was either dropped during development or lives in a separate commit. The description should be updated to remove this bullet, or the feature should be included.

2. Subtle || vs ?? inconsistency in scs_remediation_pr bodyBuilder
...(body.purl || input.purl ? { purl: (body.purl ?? input.purl) as string } : {}),
The condition uses || (falsy check) while the value uses ?? (nullish check). If body.purl is an empty string "":

Condition: "" || input.purl → truthy if input.purl is set
Value: "" ?? input.purl → "" (empty string is not nullish)
This would produce purl: "" in the request body. The operators should be consistent — either both || or both ??. Recommended fix: use || for both, so empty strings fall through to input.purl:

...(body.purl || input.purl ? { purl: (body.purl || input.purl) as string } : {}),
Same issue applies to target_version on the next line.

3. No unit tests for coerceRecord
The new coerceRecord utility in src/utils/type-guards.ts is used across all 6 tool handlers but has zero dedicated test coverage. Edge cases to test:

coerceRecord('{"a":1}') → { a: 1 }
coerceRecord('not json') → undefined
coerceRecord('[]') → undefined (array, not record)
coerceRecord('null') → undefined
coerceRecord(42) → undefined
coerceRecord({ a: 1 }) → { a: 1 } (passthrough)
coerceRecord(undefined) → undefined
4. No unit tests for the ELK→Mongo fallback logic in Registry.dispatch
The fallback mechanism in src/registry/index.ts is the most complex new logic path, but it has no dedicated unit test. The test file only adjusts existing 404 tests to add extra mock rejections. There should be explicit tests for:

ELK succeeds → _data_source: "elasticsearch" on response
ELK fails with 500 → fallback to Mongo succeeds → _data_source: "mongodb"
ELK fails with 401 → no fallback, throws immediately
ELK fails with 403 → no fallback, throws immediately
ELK fails with non-HarnessApiError (e.g. network error) → no fallback, throws
Both ELK and Mongo fail → Mongo error propagates
5. No unit tests for new resources (scs_component_vulnerability, scs_component_search)
These new resource definitions have no test coverage for:

Resource registration and field validation
pathBuilder logic (account-scoped vs artifact-scoped paths)
responseExtractor behavior (empty results, non-empty results, _reminder injection)
searchAliases, relatedResources, listFilterFields
This is a gap compared to existing SCS resources which have thorough test coverage in tests/registry/scs.test.ts.

6. No unit tests for new custom extractors
Nine new extractors are introduced but none are individually tested:

artifactSecurityListExtract
artifactSourceListExtract
componentDependenciesExtract
componentVulnerabilityExtract
componentRemediationExtract
projectSecurityOverviewExtract
bomViolationListExtract
codeRepoListExtract
artifactComponentListExtract

The existing tests were only patched to accommodate changed array lengths (e.g., expect(result).toHaveLength(2) instead of 1), but the anti-fabrication hints themselves aren't validated.

**Minor Observations (Non-Blocking)**
Anti-fabrication hints are verbose. Some _reminder strings exceed 100 words (e.g., componentVulnerabilityExtract empty case is ~75 words). Every list response now carries this overhead. This is a deliberate trade-off against hallucination, and it's defensible, but worth monitoring token cost impact. Consider whether shorter hints could achieve the same effect.

_data_source mutation. The result object is mutated in-place via (result as Record<string, unknown>)._data_source = dataSource. This works because extractors create fresh objects, but it's fragile if an extractor ever returns a frozen or shared reference. Adding a defensive Object.assign({}, result, { _data_source: dataSource }) would be safer.

ELK fallback retries on all 4xx (except 401/403). This is very permissive — a 400 "Bad Request" that is not ES-specific would still trigger a Mongo retry. The comment acknowledges this ("a 400 may be ES-specific"), which is fair for the SCS context, but other consumers of elkFallback should be aware of this behavior.

scs_component_search uses limit: "20" while all others use "10". The test was properly updated to handle this exception, which is good.

**What's Good**
coerceRecord is a smart defensive fix. LLMs serializing params as JSON strings is a real-world failure mode, and this transparent parsing is elegant.

The ELK→Mongo fallback is well-designed at the architecture level. The elkFallback flag on EndpointSpec keeps it declarative, and the fallback logic in Registry.dispatch is centralized.

Anti-fabrication extractors address a real problem. LLM hallucination of CVEs, versions, and dependency relationships is dangerous in a security context.

The pathBuilder on scs_component_vulnerability correctly supports dual-mode (account vs artifact scoped).
BOM violation field list expansion (adding snake_case variants) is a good defensive measure against API response format variations.

Test updates are correct — existing tests properly adjusted for new array shapes.

**Verdict**
The code quality and architecture are solid. The main concern is test coverage: the PR introduces significant new logic paths (ELK fallback, 9 custom extractors, 2 new resources, coerceRecord) with minimal test additions. The existing test patches are correct but insufficient — they adjust for side effects rather than validating the new behavior.

Recommendation: Request tests for coerceRecord, the ELK fallback mechanism, and at minimum the scs_component_vulnerability resource registration (following the pattern already established for other SCS resources in scs.test.ts). The HARNESS_SCS_BASE_URL mention in the PR description should be removed or the feature added.

thisrohangupta (approved):

PR #76, opened 2026-04-21
feat: [SSCA-6294]: preflight hooks, anti-fabrication guards, and scopeParams for SCS resources

mohit-agarwal-harness:
## What
- Add `PreflightContext` interface and `preflight` hook to `EndpointSpec` — enables server-side invariant checks (e.g. duplicate detection) before create/update requests are dispatched
- Add `scopeParams` to `scs_auto_pr_config` so org/project are injected as query-scoped env-var fallbacks instead of path params (aligns with the actual `/v1/ssca-config/auto-pr-config` endpoint which uses account-header scoping)
- Add anti-fabrication `_reminder` field to `scs_project_security_overview` extractor output, instructing the LLM to report only API-returned numbers without computing derived metrics
- Wire `preflight` support in the registry dispatcher (`src/registry/index.ts`) and export the new type from `src/index.ts`

## Why
The LLM was fabricating percentages, risk scores, and trend data not present in API responses. The `_reminder` field in extractor output and the preflight hook infrastructure address this by constraining agent behavior at the tool layer. The `scopeParams` fix resolves 404s on the auto-PR config endpoint when org/project were incorrectly templated into the URL path.

## References
- **Jira:** [SSCA-6294](https://harness.atlassian.net/browse/SSCA-6294)

### Repo Checks
- **Typecheck:** passed
- **Build:** passed  
- **Unit tests:** 943/943 passed (0 failures)

### SCS Eval Suite (latest run, Apr 22 2026)

| Suite | Cases | Passed | Failed | Pass Rate | Duration |
|-------|-------|--------|--------|-----------|----------|
| Single-turn (48 queries) | 48 | 47 | 1 | 98% | 924s |
| Multi-turn (19 conversations) | 19 | 19 | 0 | 100% | 720s |
| **Combined** | **67** | **66** | **1** | **98.5%** | **1644s** |

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Documentation
- [ ] Other

## Checklist
- [x] Tests pass
- [x] Typecheck passes

[SSCA-6294]: https://harness.atlassian.net/browse/SSCA-6294?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -64,6 +64,8 @@ function createHarnessServer(config: Config): McpServer {
         "DISCOVERY: Use harness_describe() to list all resource types, or harness_describe(resource_type='<type>') for operations and fields.",
         "SCHEMA: Use harness_schema(resource_type='<type>') to fetch the exact JSON Schema for create/update body payloads.",
         "",
+        "ENTITY SELECTION: When the user references an ordinal ('first repo', 'second artifact', 'latest execution'), pick the item at that index from the list response (0 = first, -1 = last). Do NOT substitute a different item or pick by name unless the user asks by name. If the list is empty, say so — never guess an ID.",
+        "",
         "PR RESOURCES: pull_request, pr_comment, pr_activity, pr_reviewer, pr_check — all accept URL or explicit repo_id + pr_number.",
         ...(config.HARNESS_PIPELINE_VERSION === "1"
           ? ["", "PIPELINE VERSION: This account uses v1 pipelines. Use resource_type='pipeline_v1' for all pipeline operations."]
--- a/src/registry/index.ts
+++ b/src/registry/index.ts
@@ -283,6 +283,11 @@ export class Registry {
     input: Record<string, unknown>,
     signal?: AbortSignal,
   ): Promise<unknown> {
+    // Run preflight hook (e.g. duplicate-check before create) before hitting the API.
+    if (spec.preflight) {
+      await spec.preflight({ client, input, registry: this, signal });
+    }
+
     // Build path with substitutions (or pathBuilder when present)
     let path: string;
     if (spec.pathBuilder) {
--- a/src/registry/toolsets/scs.ts
+++ b/src/registry/toolsets/scs.ts
@@ -220,6 +220,29 @@ const COMPONENT_DRIFT_LIST_FIELDS = [
   "status", "old_component", "new_component",
 ];
 
+const REMEDIATION_PR_LIST_FIELDS = [
+  "id", "purl", "package_name", "current_version", "target_version",
+  "pr_url", "pr_number", "pr_status", "repo_name",
+  "base_branch", "remediation_branch",
+  "created_at", "updated_at", "trigger_type", "created_by",
+];
+
+/**
+ * Custom extractor for scs_remediation_pr list responses.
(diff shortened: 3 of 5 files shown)


thisrohangupta:
Review concerns from PR #76:
 
1. **High:** Potential duplicate-PR miss due to lossy PURL normalization in preflight (`src/registry/toolsets/scs.ts`, around `normalize` in create preflight). Splitting on `@` can mis-handle valid PURLs containing `@` in namespace/qualifiers.
2. **Medium:** Duplicate check is fail-open if list call errors (`src/registry/toolsets/scs.ts`, preflight `catch { return; }`), which can bypass invariant enforcement during transient failures.
3. **Medium:** Missing targeted tests for newly added preflight and remediation list extractor behavior (`tests/registry/scs.test.ts` has basic shape checks but no conflict/no-conflict/list-failure assertions).
 
Open questions:
- Confirm actual remediation list field names for PR metadata (`pr_url/pr_number/pr_status` vs `pull_request_url/pull_request_number/status`) to ensure conflict diagnostics are populated correctly.
- Confirm whether fail-open on preflight list failure is intentional product behavior or should be fail-closed for duplicate prevention.

thisrohangupta:
Review findings
Important — duplicate detection is too broad and can block legitimate future PRs
src/registry/toolsets/scs.ts:910-918
The new preflight normalizes PURLs with split("@")[0], which drops the version and then treats any existing PR for the same package as a duplicate. It also does not consider pr_status, current_version, or target_version. If an artifact already had a merged/closed remediation PR for pkg:npm/express@4.18.0, a later attempt to create a new remediation PR for a different express version on the same artifact would still be blocked.
Important — the duplicate-block error message references field names that the extractor does not preserve
src/registry/toolsets/scs.ts:223-227, src/registry/toolsets/scs.ts:921-924
remediationPrListExtract keeps pr_url, pr_number, and pr_status, but the preflight error tries to read pull_request_url, pull_request_number, and status. In the conflict case, the code will likely block creation without surfacing the conflicting PR URL/number/status that the message claims it can provide, making the operator workflow much harder.
Moderate — the “server-side invariant” only checks the first 100 PRs
src/registry/toolsets/scs.ts:880-886
The preflight list call is hard-coded to { page: 0, size: 100 } and never paginates. For artifacts with more than 100 remediation PRs, duplicates beyond the first page are invisible, so the new duplicate-prevention logic is incomplete and can still allow the exact class of duplicate it is meant to prevent.
Test gaps
tests/registry/scs.test.ts:582-637 covers static metadata for scs_remediation_pr, but there is no test for the new duplicate-preflight behavior.
I did not find coverage proving that EndpointSpec.preflight in src/registry/index.ts short-circuits the outbound request when it throws.
I did not find a test for remediationPrListExtract verifying that { items: [...] } is flattened and that pr_url / pr_number / pr_status survive list handling.
I did not find a test for the scs_auto_pr_config path/scope change to /v1/ssca-config/auto-pr-config with query-scoped org/project params.

thisrohangupta (approved):

PR #82, opened 2026-04-24
fix: [SSCA-6347]: correct policy_set type filter for SBOM enforcement

mohit-agarwal-harness:
### What
Fix the governance MCP toolset's `policy_set.list` so the agent can enumerate SBOM policy sets. Previously the `policy_set` resource told the agent to filter SBOM enforcement policy sets with `type='ssca_enforcement'`, but the backend (`GET /pm/api/v1/policysets`) stores SBOM policy sets under `type='sbom'`. Every call returned an empty list and AIDA incorrectly answered "no SSCA enforcement policy sets configured at any scope" on projects like `SSCA / SSCA_Sanity_Automation` that have 10 configured sets (including `opa_policy_set_for_at`).

Changes:
- `policy_set` resource: description now prescribes `type='sbom'`, warns against both wrong values (`sbom_enforcement` and `ssca_enforcement`), documents the policy-vs-policy_set type asymmetry, and names `action='onstep'` for the SBOM enforcement step.
- `policy` resource: expose the `type` query param and `listFilterField` so `type='sbom_enforcement'` round-trips to `/pm/api/v1/policies`.
- `scs_bom_violation` ([scs.ts](cci:7://file:///Users/thecopyninja/IdeaProjects/mcp-server/src/registry/toolsets/scs.ts:0:0-0:0)): add `policy_set` as a sibling `relatedResource` with explicit type-filter guidance so the agent can walk the `violation → policy_set → policy` chain.
- Governance tests: remove the obsolete assertion that baked the bug into the contract; add regression coverage for the corrected type guidance, the new policy `type` query param, the asymmetry banner, and the `scs_bom_violation` back-edge.
- SCS tests: add end-to-end regression for the `scs_bom_violation ↔ policy_set ↔ policy` chain.

### Why
Closes the last SSCA-6347 AC blocker: "Integration/regression test covers the governance cross-toolset chain: SCS artifact violation → governance policy set → governance policy." Without this fix the demo prompt "show me the OPA policy sets used by SSCA enforcement" returns an empty list and the agent fabricates a "none configured" response even when sets exist.

### References
- **Jira:** [SSCA-6347](https://harness.atlassian.net/browse/SSCA-6347)
- **PRD:** N/A

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Documentation
- [ ] Other

## Checklist
- [x] Tests pass
- [x] Typecheck passes

[SSCA-6347]: https://harness.atlassian.net/browse/SSCA-6347?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/registry/toolsets/governance.ts
+++ b/src/registry/toolsets/governance.ts
@@ -30,8 +30,8 @@ const policySetCreateSchema: BodySchema = {
   fields: [
     { name: "identifier", type: "string", required: true, description: "Unique identifier for the policy set" },
     { name: "name", type: "string", required: true, description: "Display name" },
-    { name: "action", type: "string", required: true, description: "Enforcement action: onrun, onsave, onpush, etc." },
-    { name: "type", type: "string", required: true, description: "Entity type: pipeline, connector, service, environment, etc." },
+    { name: "action", type: "string", required: true, description: "Enforcement action: onrun, onsave, onpush, onstep (SBOM enforcement), onstepstart." },
+    { name: "type", type: "string", required: true, description: "Entity type. For SBOM/SSCA enforcement use 'sbom' (the policies inside this set use 'sbom_enforcement'). Other values: pipeline, connector, service, environment, secret, template, infrastructure." },
     { name: "enabled", type: "boolean", required: true, description: "Whether the policy set is enabled" },
     { name: "description", type: "string", required: false, description: "Description of the policy set" },
     { name: "policies", type: "array", required: false, description: "Policies to include: [{ identifier, severity }]", itemType: "{ identifier: string, severity: 'warning' | 'error' }" },
@@ -43,8 +43,8 @@ const policySetUpdateSchema: BodySchema = {
   description: "OPA policy set update",
   fields: [
     { name: "name", type: "string", required: false, description: "Updated display name" },
-    { name: "action", type: "string", required: false, description: "Updated enforcement action" },
-    { name: "type", type: "string", required: false, description: "Updated entity type" },
+    { name: "action", type: "string", required: false, description: "Updated enforcement action: onrun, onsave, onpush, onstep (SBOM enforcement), onstepstart." },
+    { name: "type", type: "string", required: false, description: "Updated entity type. For SBOM/SSCA enforcement use 'sbom' (policies inside use 'sbom_enforcement'). Other values: pipeline, connector, service, environment, secret, template, infrastructure." },
     { name: "enabled", type: "boolean", required: false, description: "Enable or disable the policy set" },
     { name: "description", type: "string", required: false, description: "Updated description" },
     { name: "policies", type: "array", required: false, description: "Updated policies: [{ identifier, severity }]", itemType: "{ identifier: string, severity: 'warning' | 'error' }" },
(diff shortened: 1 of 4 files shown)


thisrohangupta:
Review — SSCA-6347 policy_set type filter fix
I pulled the branch, ran pnpm typecheck and pnpm test (1001/1001 passing), and reviewed against our standards (CLAUDE.md, AGENTS.md, CONTRIBUTING.md). Overall this is a solid bug fix — scoped correctly, test-first, and the fix lives in the right layer (declarative ResourceDefinition metadata, not code), which matches our "prefer data over prose" / "put guidance on EndpointSpec" rule.

What's good
Right layer of the fix. The asymmetry is captured on the policy_set resource description, listFilterFields[type], and relatedResources — all surfaced automatically via harness_describe. No bloat added to the server instructions block — matches the anti-bloat rules.

policy list now actually accepts type. The missing queryParams.type wiring on /pm/api/v1/policies was the real blocker for the policy leg of the chain — glad that's included, not just docs.

Regression coverage is tight. Replacing the old ssca_enforcement assertion with an active warning test (Do NOT pass type='sbom_enforcement' or type='ssca_enforcement') inverts the contract so this specific bug cannot silently regress. The end-to-end violation ↔ policy_set ↔ policy structural test is exactly the SSCA-6347 AC.

Cross-toolset back-edge. Adding scs_bom_violation as a sibling on policy_set, and policy_set/policy as siblings on scs_bom_violation, closes the discovery loop in both directions.

Naming/scope. PR is narrowly scoped to the one bug ("One concern per PR" from CONTRIBUTING.md). No drive-by refactors.


**Suggestions**
policy_evaluation.type filter is stale in the same way. In governance.ts around line 234 it still says "Filter by entity type (pipeline, connector, service, etc.)" with no mention of sbom / sbom_enforcement. If an agent tries to filter evaluations for SBOM policy sets they'll hit a variant of the same failure mode. Low-risk one-line fix that belongs in this PR since it's the same bug family.

policySetCreateSchema.type / policySetUpdateSchema.type body-schema descriptions (lines 34, 47) still say "Entity type: pipeline, connector, service, environment, etc." with no mention of sbom. If an agent tries to create an SBOM policy set from this schema alone via harness_describe, they won't know sbom is the correct value. Worth a matching one-liner so create/update output agrees with the list guidance.
Consider enum on the type filters. FilterFieldSpec already supports enum: string[] (see pull-requests.ts, pipelines.ts). An explicit enum would make the valid-value set machine-readable rather than prose-only, making "unknown values return empty list" structurally impossible to hit:

{ name: "type", description: "...", enum: ["sbom", "pipeline", "connector", "service", "environment", "secret", "template", "infrastructure"] }
Same for policy.type with sbom_enforcement. The prose warning is still useful — enum just prevents the LLM from getting close to the wrong value.
action filter enum. While touching this, action is a small closed set (onrun, onsave, onpush, onstep, onstepstart) — an enum there would fully document the onstep ↔ SBOM-enforcement binding you already mention in the description.

Minor: description length. The policy_set.description is now ~8 lines of dense prose. It's all load-bearing, but if it grows further consider splitting the asymmetry warning into diagnosticHint — that field exists specifically for "if you hit an empty result, try X" guidance, and would keep the top-of-resource description focused on "what is this." Not a blocker.
Standards checklist

pnpm typecheck
passes

pnpm test
passes (1001/1001)
Declarative data change, not code — matches "Prefer data over prose"
No changes to server
instructions
block — matches anti-bloat rules
Zod 4 / stderr-logging rules N/A (no handler code touched)
One concern per PR

Regression tests added and old buggy assertion replaced (not just supplemented)


Verdict: LGTM once (1) and (2) are addressed — those are the same asymmetry leaking into two other places. (3), (4), (5) are optional polish.

PR #83, opened 2026-04-24
updated the existing CRUD operation on the custom agent

patelraj0602:
## Type of Change
- [ ] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Documentation
- [ ] Other

## Checklist
- [ ] Tests pass
- [ ] Typecheck passes


thisrohangupta:
What the PR does


Rewrites the create-agent prompt (src/prompts/create-agent.ts) from a long agent-pipeline story (stages, clone, harness_schema + path=…, secrets, raw MCP URLs, allowed_tools, chaining, custom skills) to a narrower “standalone agent” model: agent.step.run, harness-ai-agent image, llmConnector / mcpConnectors, mcp_format: harness, and stricter “do not add clone/platform/stages/allowed_tools” rules.
Renames the toolset agent-pipelines → agents, swaps the import in src/registry/index.ts, updates ToolsetName in src/registry/types.ts, and edits src/registry/toolsets/agents.ts (same logical file as today’s agent-pipelines.ts in this workspace).

Removes the agent_run resource and the commented execute wiring; drops executeHint on agent.
Where it aligns with repo standards

Registry pattern is preserved: one ToolsetDefinition, resourceType, operations, bodySchema style matches the rest of the codebase.

Zod v4 usage in the prompt registration is unchanged in spirit (the PR does not introduce the bad .describe()-before-.optional() pattern on new fields).
Slimmer prompt can be good for tokens and for matching a single supported product shape if the platform really moved to the connector-driven template the PR documents.

Gaps and risks vs repo standards and maintainability

Process / hygiene
The PR body leaves the template checkboxes unchecked (“Tests pass”, “Typecheck passes”), and the CLA bot reports CLA not signed. For an org repo, that alone blocks merge regardless of code quality.

CLAUDE.md / AGENTS.md philosophy
Those docs push structured discovery (harness_describe, schema paths, metadata on resources) instead of huge prose. This PR removes the explicit harness_schema(resource_type="agent-pipeline", path=…) guidance from the prompt. If the API still accepts the richer agent-pipeline YAML, the prompt now actively discourages patterns that may still be valid—that is a product correctness question reviewers must confirm with the Agents API team.

Metadata consistency
After the prompt says no clone / no multi-stage, the toolset description still mentions cloning repos and generic MCP in the GitHub patch. That mismatch will confuse both humans and models using harness_describe.

wiki field semantics
The patch changes the create/update body schema copy from markdown documentation to “URL to documentation”. If the API still stores markdown (as before), bodySchema would be misleading—that violates the repo’s own rule that agent guidance should stay accurate because harness_describe surfaces it.

Breaking surface for operators
Renaming the toolset to agents breaks HARNESS_TOOLSETS (and any docs) that reference agent-pipelines. There is no deprecation alias in the patch snippet—worth calling out in review unless the team explicitly wants a hard rename.

Removing agent_run (and execute stubs)
If the backend still exposes run listing or execute, removing agent_run is a capability regression with no transition note in README or tests. If execute was intentionally not shipped, removing commented code is fine, but then README / resource tables (this repo lists agent_run and agent-pipelines) need updating in the same PR.

Tests
There are no tests tied to this toolset in-tree today, so CI might still go green, but the change still does not add coverage for the rename or for the new prompt contract—weak against regressions.

Verdict
Conceptually the PR looks like a deliberate product alignment (prompt + toolset naming + slimmer CRUD surface) rather than a style-only edit. As a repo-standard PR, it is incomplete until: CLA + checklist, docs/env (HARNESS_TOOLSETS) updated for agents, README resource table fixed if agent_run disappears, and toolset/prompt/bodySchema text reconciled so nothing contradicts the real API. I would also confirm with Harness whether dropping agent-pipeline / schema guidance is correct before approving, because that is the highest-impact behavioral change for agents using the MCP.

thisrohangupta (approved):

PR #107, opened 2026-05-02
feat: add OperationPolicy contract to registry — risk-gated confirmation

sunilgattupalle:
## Summary

- Introduces `OperationPolicy` (`risk: RiskLevel`, `retryPolicy: RetryPolicy`) as a **required** field on every `EndpointSpec`, replacing the ad-hoc `blockWithoutConfirmation` boolean
- Tool handlers (`harness_create`, `harness_update`, `harness_execute`) now derive `destructive` from `isBlockingRisk(operationPolicy.risk)` — high-risk execute actions (pipeline.run, gitops.sync, FF.kill, etc.) **block on clients without MCP elicitation**
- All ~500 endpoint specs across 31 toolset files backfilled with correct risk/retry classification
- Types `PreflightContext` with structural interfaces (`HarnessClientInterface`, `RegistryDispatchInterface`) instead of `unknown`

Implements [Spec 001: Registry as Contract](specs/001-registry-contract.md). Unblocks P3 (fail-closed elicitation), P4 (autonomous modes), P5 (retry policy), P10 (write quality contract).

### Core types (`src/registry/types.ts`)
- `RiskLevel`: `read | low_write | medium_write | high_write | destructive`
- `RetryPolicy`: `safe | idempotency_key_required | do_not_retry`
- `OperationPolicy`: `{ risk, retryPolicy }` — required on `EndpointSpec`
- `isBlockingRisk()`: returns `true` for `high_write` and `destructive`
- `HarnessClientInterface`, `RegistryDispatchInterface`: structural interfaces for typed `PreflightContext`
- **Removed**: `blockWithoutConfirmation` from `EndpointSpec`

### Tool handlers
- `harness-create.ts`: `isBlockingRisk(def.operations.create!.operationPolicy.risk)`
- `harness-update.ts`: same pattern
- `harness-execute.ts`: reads risk per-action from `actionSpec.operationPolicy.risk`

### Toolset backfill (31 files, ~500 specs)
- Group A (read-only, 8 files): all `read/safe`
- Group B (simple write, 10 files): default CRUD rules
- Group C (complex + execute, 13 files): per-action risk classification
- `repositories.ts`: removed 4 `blockWithoutConfirmation: true` usages

### Tests
- 7 structural validation tests enforcing operationPolicy on every spec
- `isBlockingRisk` unit tests (all 5 risk levels)
- Integration tests for risk-gated execute actions

### Docs
- `docs/architecture.md`: design, data flow, module dependencies, invariants
- Updated `docs/testing/repo_rule/test_plan.md` and `space_rule/test_plan.md`

## Test plan

- [x] `npx tsc --noEmit` — zero errors
- [x] `pnpm test` — 1061 tests pass (45 files)
- [x] Structural validation: every spec has operationPolicy, valid risk/retry, deletes are destructive, reads are read, no blockWithoutConfirmation
- [x] `isBlockingRisk` unit tests pass for all 5 risk levels
- [x] Integration: `harness_execute` with `high_write` action blocks without elicitation; `low_write` proceeds
- [ ] Manual: verify via MCP Inspector that pipeline.run prompts for confirmation on elicitation-capable clients


thisrohangupta:
Overall
This is a strong, directionally right change: it replaces a rarely set boolean (blockWithoutConfirmation on a handful of specs) with an explicit, typed contract on every endpoint, wires create / update / execute to isBlockingRisk → elicitation destructive, and locks the model in with structural tests plus integration tests for the new execute semantics. The split called out in docs/architecture.md—registry owns risk data; confirmViaElicitation stays a plain boolean—keeps layering clean and is easy to reason about.

What works well
Contract-first: Required operationPolicy on every EndpointSpec plus invariant tests (reads are read, deletes are destructive, valid enums, no legacy field) is the right way to make a ~500-spec backfill reviewable in practice.
Behavior change is test-backed: Updating the integration case so pipeline.run blocks when the client does not support elicitation, and adding a low_write execute path that still proceeds, directly encodes the product intent and guards regressions.
Docs + spec: specs/001-registry-contract.md and docs/architecture.md give a clear story for P3/P5 follow-ons (especially retryPolicy before the HTTP client consumes it).
Things to double-check or tighten
harness_execute default risk (from the diff):
actionSpec?.operationPolicy.risk ?? "low_write" — if anything ever slipped past the structural suite, defaulting to low_write is fail-open for safety. Tests make that unlikely; for defense in depth, a fail-closed default (e.g. treat missing policy as high_write or throw) would match the “registry as contract” narrative, even if it is “unreachable” in CI.

Static MCP tool annotations vs per-action risk: The registered tool still exposes fixed hints (e.g. destructiveHint: false on harness_execute in main). After this PR, real danger is per-action. That is fine if consumers rely on runtime elicitation only; if any client trusts static hints only, they could be misled. Worth a sentence in the PR or a small follow-up if the SDK allows dynamic hints (often it does not).

Classification drift: With ~500 annotated operations, mistakes will be statistical, not zero. The structural tests enforce shape, not semantic “is this really medium_write vs high_write?” Spot-check a few sensitive areas (GitOps sync, chaos run, FF kill, approvals) against the table in the new architecture doc.

.gitignore: Adding harness-mcp-v2-quality-review.md.pdf looks unrelated to the feature; I would drop it or justify it (local artifact vs something the repo should ignore).

PreflightContext typing: The PR body says structural interfaces instead of unknown; worth confirming the final diff actually replaces unknown everywhere that matters for hooks, so the story matches the code.

Verdict
Approve with minor nits: The design is coherent, the handler changes are minimal and localized, and the test + doc package matches the size of the data change. I would clear the stray .gitignore entry, consider a fail-closed default for missing execute operationPolicy, and optionally note static vs per-action MCP hints for future client UX. The unchecked manual MCP Inspector item in the test plan is reasonable to complete before merge if your bar is “touched elicitation UX.”

thisrohangupta (approved):

PR #108, opened 2026-05-02
fix: P1/P7/P8 — version unification, FME config wiring, debug log redaction

sunilgattupalle:
## Summary

Addresses three Phase 1 quality review items (stacked on #107):

- **P1 (Version unification):** MCP server metadata now reads version from `package.json` via `getVersion()` instead of hardcoded `"2.0.0"`. Single source of truth — CLI `--version` and MCP `initialize` response both reflect `package.json`.

- **P7 (Dead config):** `resolveProductBaseUrl()` now uses `HARNESS_FME_BASE_URL` from config (defaults to `https://api.split.io`) instead of a dead hardcoded const. Self-managed/staging environments can now override the FME base URL via env var.

- **P8 (Debug log redaction):** All 3 debug-level body log sites in `harness-client.ts` now run through `redactSensitiveFields()` by default. Redacts values for keys matching: `token`, `secret`, `password`, `apiKey`, `credential`, `privateKey`, `clientSecret`, `authorization`, `bearer`, `webhook`, `passphrase`, `encrypted`, `sshKey`, `accessKey`, `secretKey`. Set `HARNESS_LOG_UNSAFE_BODIES=true` to bypass for local debugging.

Also includes deferred spec 002 (TransportOverrides) for future reference — design is pre-approved but implementation deprioritized per review.

## Test plan

- [x] `pnpm typecheck` — clean
- [x] `pnpm test` — 1072/1072 pass (includes 11 new redaction tests)
- [x] Zero lint errors on changed files
- [ ] Verify `--version` CLI output matches `package.json`
- [ ] Verify MCP `initialize` response `serverInfo.version` matches `package.json`
- [ ] Verify debug logs with `LOG_LEVEL=debug` show `[REDACTED]` for sensitive fields
- [ ] Verify `HARNESS_LOG_UNSAFE_BODIES=true` bypasses redaction

### DIFF
--- a/specs/002-endpoint-spec-transport-overrides.md
+++ b/specs/002-endpoint-spec-transport-overrides.md
@@ -1,0 +1,361 @@
+# Spec 002: EndpointSpec Transport Overrides
+
+**Status:** Deferred (pre-approved pattern — implement when 8th+ transport quirk appears)
+**Author:** Architecture Review Follow-up
+**Date:** 2026-05-02
+**Depends on:** 001 (OperationPolicy)
+**Unblocks:** Cleaner extension model for future product integrations
+
+---
+
+## Problem
+
+`EndpointSpec` has accumulated 7 optional fields that serve 1–3 modules each.
+They are legitimate API-behavior quirks, but they bloat the shared contract that
+every resource definition depends on. Each new Harness product integration risks
+adding more module-specific fields:
+
+| Field | Modules that use it | Purpose |
+|---|---|---|
+| `elkFallback` | scs (5 specs) | ELK-to-Mongo retry for ssca-manager |
+| `headerBasedScoping` | sei, ai-evals, pipelines (~32 specs) | Scope via `Harness-Account` header |
+| `injectAccountInBody` | gitops, ccm (6 specs) | gRPC-gateway body injection |
+| `injectOrgQueryFallback` | platform (4 specs) | NG project POST needs org query param |
+| `pageOneIndexed` | dashboards (1 spec) | 1-based pagination shim |
+| `emptyOnErrorPatterns` | gitops (2 specs) | Swallow "no data" 500s |
+| `responseType` | dashboards (1 spec) | Binary/buffer response |
+
+Additionally, two dead fields exist today:
+
+- **`EndpointSpec.scopeParams`** — declared in the interface but never set by
+  any toolset and never read by `executeSpec`. Only `ResourceDefinition.scopeParams`
+  is used.
+- **`emptyOnErrorPatterns`** — set on 2 GitOps specs but `executeSpec` never
+  reads it. The error-swallowing behavior was never wired up.
+
+---
+
+## Design Alternatives Considered
+
+### Alternative A: Interface Extension / Inheritance
+
+```typescript
+// Core
+interface EndpointSpec { method: HttpMethod; path: string; ... }
+
+// Per-module extension
+interface SCSEndpointSpec extends EndpointSpec { elkFallback?: boolean; }
+interface SEIEndpointSpec extends EndpointSpec { headerBasedScoping?: boolean; }
+```
+
+**Why this doesn't work for us:**
+
+The fundamental constraint is that `executeSpec` is a single centralized method
+that receives `EndpointSpec` via `ResourceDefinition.operations`. If toolsets
+define extended interfaces, the dispatch layer receives the base `EndpointSpec`
+and *cannot see* extension fields without type narrowing.
+
+To make it work you'd need one of:
+
+1. **Type assertions in `executeSpec`** — e.g. `(spec as SCSEndpointSpec).elkFallback`.
+   This defeats the purpose. Dispatch still knows about every module, and now
+   with unsafe casts instead of optional typed fields.
+
+2. **Generic `ResourceDefinition<T extends EndpointSpec>`** — the generic
+   cascades to `ToolsetDefinition<T>`, the registry `Map`, every dispatch
+   overload, and every tool handler. Massive complexity for no behavioral gain.
+
+3. **Discriminated union** — `EndpointSpec = SCSEndpointSpec | SEIEndpointSpec | ...`
+   Forces every consumer to narrow the union. Fields like `headerBasedScoping`
+   span multiple products, so the discriminant doesn't align to product.
+
+**Verdict:** Extension/inheritance models solve a problem we don't have.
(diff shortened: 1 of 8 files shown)


thisrohangupta:
Overall: Approve with minor follow-ups
The PR cleanly addresses the three stated review items (P1/P7/P8). Implementation is small, focused, well-tested (11 new redaction tests, 1085/1085 passing), and pnpm typecheck is clean. Worth merging once the documentation gap and a test coverage gap are addressed.

P1 — Version unification
Clean. Exporting getVersion() from src/utils/cli.ts and using it in the McpServer ctor makes package.json the single source of truth. Verified locally: getVersion() returns 0.9.6 matching package.json. CLI --version and serverInfo.version now stay in sync automatically on every release bump.

P7 — FME base URL wiring
resolveProductBaseUrl(config, "fme") now reads from config.HARNESS_FME_BASE_URL instead of the dead hardcoded const. The HTTPS guard mirrors the one for HARNESS_BASE_URL, and the new tests in tests/config.test.ts cover the three cases (reject HTTP by default, accept HTTP when HARNESS_ALLOW_HTTP=true, accept HTTPS).

The bonus FME transport fixes in commit 17d47761 are correct and necessary:

harness-client.ts
Lines 192-198
    const isFme = options.product === "fme";
    const accountId = this.resolveAccountId();
    const headers: Record<string, string> = {
      ...(isFme ? {} : { "Harness-Account": accountId }),
      ...options.headers,
    };
harness-client.ts
Lines 459-468
    const params = new URLSearchParams();
    if (!options.headerBasedScoping && options.product !== "fme") {
      const accountId = this.resolveAccountId();
      params.set("accountIdentifier", accountId);
      // Log-service gateway expects accountID (capital ID) in query params
      if (path.includes("/log-service/")) {
        params.set("accountID", accountId);
      }
    }
Without these, every FME call would have leaked Harness-Account and accountIdentifier to Split's API, which would either be ignored or trigger validation noise.

Issue (must-fix nit): the FME transport changes have zero test coverage. tests/client/harness-client.test.ts covers baseUrl override but never asserts that product: "fme" omits Harness-Account or accountIdentifier. A regression to either branch would silently start leaking account scope to Split.io. Recommend adding two assertions in request — headers / request — URL describes:

it("omits Harness-Account header for product='fme'", async () => {
  fetchSpy.mockResolvedValue(new Response("{}", { status: 200 }));
  const client = new HarnessClient(makeConfig());
  await client.request({ path: "/internal/api/v2/x", product: "fme" });
  const headers = fetchSpy.mock.calls[0][1]?.headers as Record<string, string>;
  expect(headers["Harness-Account"]).toBeUndefined();
});
it("omits accountIdentifier query param for product='fme'", async () => {
  // assert URL has no accountIdentifier
});
Design observation (not blocking): there are now two overlapping flags — headerBasedScoping ("skip query, keep header") and product === "fme" ("skip both") — implicitly coupled in buildUrl. A future cleanup could collapse to scopingMode: "query" | "header" | "none", but that's clearly out of scope here.

P8 — Debug log redaction
src/utils/redact.ts is well-designed:

Depth cap (10) — verified that anything deeper collapses to [REDACTED] instead of leaking. Defense in depth.
Anchored key pattern (^...$) — deliberately strict to avoid over-redacting keys like tokenize / secretCount. Trade-off: custom key names like gitToken, accountSecret will slip through. The PR's choice is the safer of the two; worth a code comment explaining why.
redactJsonString parse-then-fallback — primary path parses JSON, fallback uses inline regex. Verified empirically that inline alternation precedence works correctly for compound keys (access_token, webhook_url, client-secret etc. all redact properly even though token/secret come first in the alternation).
No mutation — confirmed with the dedicated test.
The 223-line test file is solid and exercises both top-level keys, nested objects, arrays, mixed cases, kebab/snake/camel, and the depth limit.

One real gap: the call sites in harness-client.ts aren't tested for the HARNESS_LOG_UNSAFE_BODIES toggle behavior. The redaction utility is heavily tested in isolation; the integration ("when flag is true, raw body is logged; when false, redacted") relies on visual inspection. Not blocking, but a one-line assertion would close the loop.

Test 4 from my probe also surfaces a non-issue worth being aware of: stringified secrets inside JSON values (e.g. {"yamlBody": "token: secret123"}) are NOT scrubbed because the redactor short-circuits on strings. This is correct behavior for "Request body" logs (the keys are what we trust as a redaction signal), but if anyone ever logs a YAML-as-string body, secrets in those values will pass through. Probably worth a JSDoc note on redactSensitiveFields.

Documentation gap
Neither README.md nor .env.example document the two new env vars introduced in this PR:

HARNESS_FME_BASE_URL — useful for self-managed/staging FME backends
HARNESS_LOG_UNSAFE_BODIES — escape hatch for local debug
These should be added to .env.example (under "Security and transport controls" makes sense) and the env-var table in README.md so users can discover them.

Spec 002
361-line design doc for a deferred refactor lives in specs/002-endpoint-spec-transport-overrides.md. The analysis is thorough and the chosen TransportOverrides bag is the right call (I agree with the reasoning rejecting interface extension/middleware/declaration merging given that all consumption is in one executeSpec function).

Two suggestions:

Land it as a separate doc-only commit/PR rather than bundled here. PR 108's stated scope is P1/P7/P8; the spec is unrelated to those. Mixing 361 lines of design discussion with three small fixes makes the PR harder to skim.
The spec catalogues EndpointSpec.scopeParams as dead. I confirmed: it's declared in src/registry/types.ts:293 but never set or read; only ResourceDefinition.scopeParams (line 199) is used. Easy 1-line cleanup that could land independently.
Verification I ran
pnpm install --frozen-lockfile — clean
pnpm typecheck — clean
pnpm test — 1085/1085 pass (PR description says 1072/1072; the gap is from PR #107's test additions also present here)
pnpm build — clean
Manually exercised redactJsonString against ~15 inputs (kebab/snake/camel, deeply nested, alternation precedence, parse failures, truncation, empty strings, numeric values) — all behaved correctly
getVersion() returns 0.9.6 matching package.json
Recommendation
Approve once:

FME transport assertions added to tests/client/harness-client.test.ts (Harness-Account header omission, accountIdentifier query omission).
HARNESS_FME_BASE_URL and HARNESS_LOG_UNSAFE_BODIES documented in .env.example + README.md env table.
Optional follow-ups (separate PRs):

Move specs/002-endpoint-spec-transport-overrides.md to its own doc PR.
Add an integration test for the HARNESS_LOG_UNSAFE_BODIES toggle in harness-client.ts.
Clean up dead EndpointSpec.scopeParams field (1-line change).

PR #112, opened 2026-05-03
spec: visualization architecture review (006)

sunilgattupalle:
## Summary

- Adds `specs/006-visualization-review.md` documenting the current server-side SVG/PNG chart rendering pipeline, known issues, and a phased migration plan.
- The server bundles 17 SVG source files + a native `@resvg/resvg-js` dependency to render inline chart PNGs, but text labels don't render (font loading disabled), the native binary has portability issues, and the approach doesn't leverage modern MCP client rendering capabilities.
- Recommends a 4-phase migration: fix fonts (immediate) -> return structured `chart_data` alongside visuals -> deprecate `include_visual` params -> remove SVG rendering code entirely.

## Test plan

- [ ] Review spec for accuracy against current codebase
- [ ] Confirm migration phases are non-breaking until final removal
- [ ] Validate open questions around `harness_diagnose` diagrams

### DIFF
--- a/specs/006-visualization-review.md
+++ b/specs/006-visualization-review.md
@@ -33,9 +33,9 @@ The MCP server currently bundles a full SVG chart rendering pipeline and a nativ
 | `timeline.ts` | Execution timeline (step-level Gantt-style) |
 | `stage-flow.ts` | Pipeline stage flow diagram |
 | `status-summary.ts` | Status summary card |
-| `executions-timeseries.ts` | Stacked bar chart by day |
-| `list-visuals.ts` | Aggregation logic + chart dispatch for `harness_list` |
-| `mappers.ts` | Data mappers from API responses to chart data types |
+| `executions-timeseries.ts` | Stacked-bar-chart-by-day renderer used by `harness_list` (via `list-visuals.ts` for `visual_type=timeseries`) |
+| `list-visuals.ts` | Aggregation logic + chart dispatch for `harness_list` (execution-only) |
+| `mappers.ts` | Tool-handler-output → SVG-renderer-input mappers (e.g. diagnose result → execution summary, project-health output → status data) |
 | `architecture.ts` | Pipeline YAML to architecture diagram |
 
 ### Rendering pipeline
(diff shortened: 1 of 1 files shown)


thisrohangupta (approved):
Approved with corrections pushed directly to `viz-review-spec` (commit `546d2c78`).

## What I changed and why

Audited every factual claim in the spec against the current codebase via parallel verification (60 claims, 51 confirmed, 9 needed correction). The 3 inaccuracies that materially mislead a reader, plus 4 recommendation issues that would have broken the implementation PR's pre-push review, are now fixed:

**Factual corrections**
- **Entry points / scope.** Visualization is wired into **3 tools** (`harness_list`, `harness_diagnose`, `harness_status`) and registered as **7 resource types** in `src/registry/toolsets/visualizations.ts` — not "one tool, one resource type." The execution-only gate exists *inside* `list-visuals.ts`, but `harness_diagnose` covers `timeline`/`flow`/`architecture` and `harness_status` covers `health_dashboard`.
- **`executions-timeseries.ts`** is used by `harness_list` (via `list-visuals.ts:141` for `visual_type=timeseries`), not `harness_status`. Inventory + entry-point list updated.
- **`harness_status` rendering** uses `renderStatusSummarySvg` only — no timeseries today.
- **Dynamic-import behavior**: the dynamic `import("@resvg/resvg-js")` lives in `render-png.ts`; the try/catch fallback is at the tool call sites, not in the renderer. Reworded.
- **Binary size**: locally measured darwin-arm64 is ~3.4 MB, not ~5 MB. Corrected to "~3–5 MB depending on platform."

**Recommendation tightening**
- **Phase 0 swapped from `loadSystemFonts: true` → embedded Inter WOFF2 via `Resvg.fontFiles`.** The original Phase 0 contradicted the spec's own portability section (Claude Desktop / headless Docker have no system fonts to load — that's where the bug actually bites). Embedding a ~100 KB font fixes every environment.
- **Phase 3 reclassified Yes(minor) → Yes(major).** Removing `include_visual` / `visual_type` is a breaking tool-contract change — external clients pin against the schema.
- **`chart_data` example uses identifier-keyed entries** (`pipeline_id`, `pipeline_name`) instead of display-name keys — aligns with the project's "identifier-named field → stable id, not a display label" pitfall (the bot would have flagged the original on the implementation PR).
- **Phase 3 preconditions made explicit** (was an unresolved open question): `harness_diagnose` diagram replacement plan, `harness_status` health-dashboard plan, and audit-sink impact (specs 004 / 005) verified before Phase 3 can ship.

## Standards check

- ✅ No internal Jira IDs in title / body / commit messages (public OSS repo rule).
- ✅ CI green on Node 20 + 22.
- ✅ No code changes, so the per-extractor / `responseExtractor` / `listFilterFields` rules don't apply — but the spec now mandates them for the implementation PR.
- ⚠️  CLA-assistant is still pending — needs `@sunilgattupalle` to sign. Once signed, this PR is mergeable.

## What I did NOT touch

- Open Question #2 (should `harness_status` return chart data) — kept as an open question because it's a product call, not a correctness one.
- The Status / Author / Created header — left alone.

If you'd rather the recommendation rewrites land in a follow-up PR, happy to revert that portion and keep only the factual corrections in this one — just let me know.

PR #115, opened 2026-05-03
feat: registry quality, safety layer, audit sinks (P1–P10)

sunilgattupalle:
## Summary

Single squashed diff of the full P1–P10 stack onto main (51 files, +3451/−142):

- **P1/P7/P8** — Version unification via `getVersion()`, `HARNESS_FME_BASE_URL` config + HTTPS guard, debug-log redaction (JSON walk + inline fallback, kebab-case keys, `webhookUrl`/`callbackUrl` patterns), FME transport contract (skip `Harness-Account` and `accountIdentifier` for `product: "fme"`)
- **P3/P4** — Fail-closed elicitation for `medium_write`+, risk-based auto-approve (`HARNESS_AUTO_APPROVE_RISK`)
- **P5** — `retryPolicy` threaded from `OperationPolicy` into HTTP client
- **P10/P2/P6** — Structural validation with dual v0/v1 registries, execute-action `bodySchema` contract, `generate-docs.js` dynamic toolset discovery, smoke test with live `harness_describe` surface validation, docs/spec sync for `requiresConfirmation()` model
- **P9** — Pluggable audit sinks (`AuditManager` → stderr/JSONL/webhook/OTel), self-bootstrapping OTel sink with external-provider detection, in-flight webhook flush tracking, graceful shutdown across stdio/HTTP, scope-accurate and tool-attributed audit events

### Specs added
- `specs/002-endpoint-spec-transport-overrides.md`
- `specs/003-safety-layer-p3-p4-p5.md`
- `specs/004-audit-sinks.md`
- `specs/005-otel-audit-sink.md`

## Test plan
- [x] `pnpm build` — clean TypeScript compilation
- [x] `pnpm test` — 1128 tests pass
- [ ] CI pipeline passes
- [ ] Manual verification with MCP Inspector


sunilgattupalle:
## Backward Compatibility Notes

### `HARNESS_SKIP_ELICITATION` → `HARNESS_AUTO_APPROVE_RISK`

**Fully backward compatible.** Existing users with `HARNESS_SKIP_ELICITATION=true` are unaffected:

- `src/config.ts` detects the old flag and maps it to `HARNESS_AUTO_APPROVE_RISK=all` (identical behavior — all operations auto-approved).
- A deprecation warning is emitted to stderr so users know to migrate.
- If both flags are set, `HARNESS_AUTO_APPROVE_RISK` takes precedence.
- The old flag remains in `ConfigSchema` and is parsed normally — no breakage.

### Behavioral change: `medium_write` now fails closed

The main behavioral change is that `medium_write` operations (flag targeting changes, protection rule creates, experiment stops, execution interrupts) now **require confirmation** on clients without elicitation support. Previously, only `high_write` and `destructive` operations were blocked.

This is intentional (P3 safety layer), but means:
- **Claude Desktop / Windsurf** users (no elicitation support yet) will see errors on `medium_write` operations unless they set `HARNESS_AUTO_APPROVE_RISK=medium_write` or higher.
- **Cursor / VS Code Copilot / MCP Inspector** users are unaffected (elicitation works).
- **CI/CD bots** using `HARNESS_SKIP_ELICITATION=true` are unaffected (maps to `all`).

### Audit sinks (P9)

No compatibility impact. Audit sinks are opt-in via environment variables (`HARNESS_AUDIT_FILE`, `HARNESS_AUDIT_WEBHOOK_URL`, `OTEL_EXPORTER_OTLP_ENDPOINT`). The default `StderrSink` is always active but only outputs to stderr (never affects tool responses).

### OTel packages

All 4 `@opentelemetry/*` packages are `optional` peer dependencies. `pnpm install` / `npm install` will not fail if they are absent. The OTel sink gracefully disables itself when packages are missing.

sunilgattupalle:
None are merge blockers. All are valid observations but represent edge-case hardening, not security vulnerabilities or data loss risks. Here's why:

Denied writes not audited — When !elicit.proceed, the tool returns early with an error message. The operation never happened, so there's no state change to audit. Logging denied attempts is an audit completeness improvement, not a safety gap. The user/agent already sees the denial in the MCP response. Follow-up quality.

Webhook concurrent flush race — The inflightFlush single-promise tracking could miss a concurrent flush. But: the webhook sink is explicitly documented as best-effort, the race window is narrow (timer-flush must fire during an active size-triggered flush's POST), and the worst case is one batch not awaited on close. Follow-up quality.

HTTP shutdown ordering — This is about graceful shutdown edge cases during server termination. The window is seconds-long and only during process exit. stdio mode (the dominant deployment) has its own shutdown path. HTTP shutdown ordering improvements are real but don't block shipping. Follow-up quality.

OTel external-provider forceFlush — By design. When reusing an external TracerProvider, the sink doesn't own the provider and shouldn't call forceFlush() on it — that's the external owner's responsibility. The sink correctly only flushes its own standalone provider. Not a bug.

Elicitation body preview leaking secrets — The bodyPreview is shown via MCP elicitation to the local user who is already providing the body and has the API key. It's not logged to stderr, not sent to a webhook, not written to audit. Redacting the preview would make confirmation unusable ("approve this operation but I can't show you what it does"). By design — follow-up at most to redact specific known-secret fields from the preview while keeping the rest readable.

Recommendation: Merge PR #115 as-is. File the denied-write audit gap (finding 1) and webhook flush race (finding 2) as follow-up issues — they're the most actionable of the five.

sunilgattupalle:
## Review comment responses

Triaging all 22 review comments. Many are duplicates across 3 review runs against different force-push versions. Grouping by topic:

---

### 1. Smoke test doesn't boot MCP server (comments on `scripts/smoke-test.js`)
**IDs**: 3178504689, 3178512625, 3178513761

**Status: Already fixed.** The smoke test on current HEAD (`50a26c1`) boots a real MCP server via `spawn()`, sends `initialize` → `notifications/initialized` → `tools/list` → `harness_describe` over stdio JSON-RPC, and validates tool/resource counts. See `checkServerStartup()` at line 144. This was fixed in the PR110 review round and merged via `phase3-p10-p2-p6-discipline`.

---

### 2. Redaction fallback for non-JSON bodies (`src/utils/redact.ts`)
**IDs**: 3178502892, 3178512618, 3178513749, 3178527001

**Status: Already fixed.** The `INLINE_SECRET_PATTERN` now uses the full `SENSITIVE_KEYS` list (including `webhook[_-]?url`, `callback[_-]?url`, kebab-case variants) and handles both `key: "value"` and `key=value` patterns. The YAML block scalar edge case (multiline `privateKey: |`) is acknowledged as a limitation of regex-based scrubbing — a full YAML parser would be needed, which is out of scope for a debug log fallback path. The primary path (JSON parse → `redactSensitiveFields`) handles structured payloads correctly.

**Follow-up potential**: YAML-aware redaction for block scalars.

---

### 3. FME HTTPS guard (`src/config.ts`)
**ID**: 3178513754

**Status: Already fixed.** Both `HARNESS_FME_BASE_URL` and `HARNESS_AUDIT_WEBHOOK_URL` now have HTTPS guards respecting `HARNESS_ALLOW_HTTP`, matching the existing `HARNESS_BASE_URL` pattern. See config.ts lines 106-123.

---

### 4. Elicitation capability detection (`src/utils/elicitation.ts`)
**IDs**: 3178499943, 3178512623

**Status: Already fixed.** `clientSupportsElicitation()` now checks `!!caps?.elicitation` (not `!!caps?.elicitation?.form`). The concern about URL-only elicitation (`{ elicitation: { url: {} } }`) is valid in theory but not in practice — no current MCP client declares URL-only elicitation, and our server only uses form-based elicitation. If a client declares URL-only, we'd attempt a form request which would fail gracefully (the client returns an error, we treat it as "not supported"). **Not a blocker.**

---

### 5. Webhook flush race / concurrent inflight tracking (`src/audit/sinks/webhook.ts`)
**IDs**: 3178502890, 3178527004

**Status: Acknowledged, not a blocker.** The webhook sink is explicitly best-effort. The concurrent flush race window is narrow (timer flush must fire during an active size-triggered flush POST). Worst case: one batch not awaited on close. The `inflightFlush` tracking + `close()` drain is a significant improvement over the pre-PR state (no tracking at all).

**Follow-up**: Serialize flushes with a queue or mutex.

---

### 6. StderrSink subject to LOG_LEVEL (`src/audit/sinks/stderr.ts`)
**IDs**: 3178502889, 3178527002

**Status: By design.** The stderr sink uses the structured logger which respects `LOG_LEVEL`. At `LOG_LEVEL=warn`, audit events (logged at `info` level) would be suppressed. This is intentional — operators who set `LOG_LEVEL=warn` are explicitly choosing to reduce output. The JSONL file sink and webhook sink are not affected by `LOG_LEVEL` and serve as the durable audit trail. Stderr is convenience, not the primary audit mechanism.

---

### 7. Crash handlers don't drain audit (`src/index.ts`)
**IDs**: 3178504686, 3178512620

**Status: Acknowledged, not a blocker.** `unhandledRejection` and `uncaughtException` handlers call `process.exit(1)` without draining audit sinks. This is standard Node.js practice — in a crash scenario, attempting async I/O (flush to webhook, write to file) is unreliable and risks hanging the process. The normal shutdown paths (SIGINT, SIGTERM, stdin EOF, stdout EPIPE) all properly drain audit. Crash = best-effort loss is acceptable.

---

### 8. HTTP shutdown doesn't reject new requests during drain (`src/index.ts`)
**IDs**: 3178513758

**Status: Acknowledged, follow-up.** The HTTP shutdown calls `server.close()` which stops accepting new connections, but existing keep-alive SSE streams on `/mcp` continue until drained. A 503 middleware for in-progress requests is a hardening improvement but not a blocker — the audit manager is shared and closed once after all sessions drain.

---

### 9. `http_path` in audit is static, not resolved (`src/registry/index.ts`)
**ID**: 3178504687

**Status: By design.** `http_path` records the endpoint template (e.g., `/ng/api/projects/{projectId}`), not the resolved URL. This is intentional — audit records should show the *operation pattern*, not the specific instance. The resolved URL with actual IDs is sensitive and would increase audit storage. The `resource_id`, `org_id`, and `project_id` fields in the audit event provide the specific context.

---

### 10. Request body leak in debug logs (`src/client/harness-client.ts`)
**ID**: 3178499944

**Status: Mitigated.** Debug body logging goes through `redactJsonString()` which handles JSON payloads (the vast majority). Raw YAML bodies are truncated at 1000 chars and scrubbed via `INLINE_SECRET_PATTERN`. The `HARNESS_LOG_UNSAFE_BODIES` toggle exists as an explicit opt-in for full body logging in local dev only. This is adequate for a debug-level log that requires `LOG_LEVEL=debug` to activate.

---

### 11. `execution.interrupt` risk level change (`src/registry/toolsets/pipelines.ts`)
**ID**: 3178520672

**Status: Intentional.** `execution.interrupt` was changed from `medium_write` to `low_write` because interrupting an already-running execution is a corrective action, not a destructive one. It's analogous to Ctrl+C — you're stopping something, not deleting data. The operation is idempotent (interrupting an already-interrupted execution is a no-op). If this needs to be revisited, it's a risk classification discussion, not a code bug.

---

### 12. Structural validation test scope (`tests/registry/structural-validation.test.ts`)
**IDs**: 3178520673, 3178527005

**Status: Acknowledged, follow-up.** The `fullRegistry` helpers feed the new write-quality assertions (bodySchema, actionDescription on risky operations). The existing invariants (type naming, description presence) still use the default registry. Extending all invariants to the full registry is a test improvement, not a blocker.

thisrohangupta (approved):
Reviewed PR 115. CI is green, and the advertised fixes for retryPolicy, FME base URL/scoping, redaction fallback, and medium-write elicitation are present.

Findings I’d address before merge:
• src/tools/harness-create.ts, src/tools/harness-update.ts, src/tools/harness-execute.ts, src/tools/harness-delete.ts: blocked/declined confirmation attempts return before registry dispatch, so security-relevant denied writes are not audited.
• src/audit/sinks/webhook.ts: concurrent flushes can overwrite the tracked in-flight promise; shutdown may miss an active POST. Final close can also leave failed batches requeued with no timer left.
• src/index.ts: HTTP shutdown destroys sessions and closes audit sinks before tracking/draining active handlers; late 503 middleware does not reliably gate existing /mcp keep-alive requests.
• src/audit/sinks/otel.ts: external-provider mode can create standalone spans but has no provider reference to forceFlush() them on close.
• src/tools/harness-create.ts and src/tools/harness-update.ts: elicitation body previews use raw body text/JSON, which can expose secrets despite the new redaction utility.

No code changes made from this automation run; this was a review request. Local targeted tests could not run because this image lacks node/pnpm, so I relied on source review plus the PR’s green CI signal.

PR #154, opened 2026-05-08
feat: support for global templates list and get

yogesh-chauhan:
## Summary

Adds support for fetching global (Harness-provided) templates via `harness_get` and `harness_list`. Also fixes TypeScript type errors introduced when `HARNESS_ORG` was made optional, and ensures `routingId` is always sent alongside `accountIdentifier` in API requests.

## What Changed

- **`harness_get` — global template support**: passing `params: { global_template: true }` on a template get now overrides `accountIdentifier` to `__GLOBAL_TEMPLATES_ACCOUNT_ID__`, enabling fetching of Harness-managed global templates
- **`templates.ts` — get op**: added `account_id → accountIdentifier` query param mapping so the override flows through cleanly; `global_template` remains a boolean at the call site
- **`templates.ts` — list op**: added `global_templates`, `metadata_only` filter fields and `pathBuilder` to route to `/list-metadata` endpoint; default `template_list_type` changed from `All` to `LastUpdated`
- **`harness-client.ts`**: `routingId` is now always sent alongside `accountIdentifier` in every request (set to the same resolved account ID)
- **`registry/index.ts`**: `orgId` getter widened to `string | undefined` to match the config type; audit events now resolve `http_path` via `pathBuilder` when present
- **`config.ts`**: removed the hardcoded `"default"` fallback for `HARNESS_ORG` — it is now truly optional
- **`harness-status.ts` / `harness-execute.ts`**: fixed type errors from optional `HARNESS_ORG` — status falls back to `""`, execute returns a clear error if org is missing when materializing input sets

## Why

Global templates (e.g. `helmDeployAction`) live under `__GLOBAL_TEMPLATES_ACCOUNT_ID__` and cannot be fetched with the caller's own account ID. The Harness API requires the `accountIdentifier` to be set to the global sentinel value while `routingId` carries the real account for request routing. Previously there was no way to do this via the MCP tool.

The `HARNESS_ORG` change and associated type fixes reflect that org is not a required field for all deployments — enforcing `"default"` as a fallback was incorrect and caused TypeScript errors once the type was properly widened.

## Type of Change
- [ ] Bug fix
- [x] New feature
- [ ] Refactor
- [ ] Documentation
- [ ] Other

## Checklist
- [x] Tests pass
- [x] Typecheck passes

### DIFF
--- a/src/registry/index.ts
+++ b/src/registry/index.ts
@@ -218,7 +218,7 @@ export class Registry {
     return new Set(valid);
   }
 
-  get orgId(): string { return this.config.HARNESS_ORG; }
+  get orgId(): string | undefined { return this.config.HARNESS_ORG; }
   get projectId(): string | undefined { return this.config.HARNESS_PROJECT; }
 
   /** Get a resource definition by type, or throw. */
@@ -390,6 +390,11 @@ export class Registry {
   ): void {
     if (!this.auditManager) return;
 
+    // Resolve path using pathBuilder if present, otherwise use static path
+    const resolvedPath = spec.pathBuilder
+      ? spec.pathBuilder(input, { HARNESS_ACCOUNT_ID: this.getAccountId(), HARNESS_ORG: this.config.HARNESS_ORG, HARNESS_PROJECT: this.config.HARNESS_PROJECT })
+      : spec.path;
+
     const event: AuditEvent = {
       event_id: randomUUID(),
       timestamp: new Date().toISOString(),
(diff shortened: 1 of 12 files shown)


src/registry/index.ts
```diff
@@ -390,6 +390,11 @@ export class Registry {
   ): void {
     if (!this.auditManager) return;
 
+    // Resolve path using pathBuilder if present, otherwise use static path
+    const resolvedPath = spec.pathBuilder
+      ? spec.pathBuilder(input, { HARNESS_ACCOUNT_ID: this.getAccountId(), HARNESS_ORG: this.config.HARNESS_ORG, HARNESS_PROJECT: this.config.HARNESS_PROJECT })
+      : spec.path;
```
yogesh-chauhan, line 396: This is to make sure the audit log contains the right path ( not the hardcoded spec.path ) in cases where tools use the PathBuilder option

hingerabhinav (approved):

thisrohangupta (approved):

PR #179, opened 2026-05-12
Fix template update body and delete path

Saranya-jena:
## Summary

- **Template delete**: Fixed delete path to include `{versionLabel}` as a path parameter instead of a query parameter, which was causing backend 500 errors when deleting a specific template version.
- **Template update/create — scope support**: The v1 template API uses different URL paths per scope (account, org, project), but the toolset hardcoded only the project-scoped path. Org-scoped and account-scoped templates failed with "Missing required field project_id" before the request ever reached the API. Added a dynamic `pathBuilder` that constructs the correct URL based on which scope fields are present.
- **Template update — body fields**: Added optional snake_case fields (`identifier`, `name`, `label`, `yaml_version`, `git_details`) as pass-through fields in the `bodyBuilder`. While `template_yaml` is the only strictly required field (the API derives metadata from the YAML content), including these optional fields ensures forward compatibility and allows explicit overrides when needed.
- **Build fix**: Added missing `SchemaEntry` type export to both `src/data/schemas/index.ts` and the `sync-schemas.js` generator so the export survives re-generation.

## Root Cause

The Harness v1 template update API (`PUT /v1/orgs/{org}/projects/{project}/templates/{template}/versions/{version}`) expects a JSON body with snake_case fields matching the Go DTO (`TemplateV1CreateRequest`). The old `bodyBuilder` only sent `template_yaml`, missing required fields like `identifier`, `label` (not `versionLabel`), `name`, `yaml_version`, and `git_details`.

## Test Plan
- [x] Verified template update via Cursor MCP tool (renamed template successfully on QA)
- [x] TypeScript typecheck passes (`tsc --noEmit`)
- [ ] Verify template delete works with the new path parameter format

### DIFF
--- a/src/registry/toolsets/templates.ts
+++ b/src/registry/toolsets/templates.ts
@@ -1,5 +1,42 @@
-import type { ToolsetDefinition } from "../types.js";
+import type { ToolsetDefinition, PathBuilderConfig } from "../types.js";
 import { ngExtract, pageExtract } from "../extractors.js";
+
+/**
+ * Builds a scope-aware v1 API base path for templates.
+ * Uses ONLY explicit input fields (org_id, project_id) — never falls back to
+ * config defaults. This lets callers target org/account scope even when
+ * HARNESS_PROJECT is configured. The dispatcher still handles default scope
+ * injection for NG endpoints (list/get/delete) via query params.
+ */
+function templateV1BasePath(input: Record<string, unknown>, _config: PathBuilderConfig): string {
+  const org = input.org_id as string | undefined;
+  const project = input.project_id as string | undefined;
+  const templateId = input.template_id as string;
+  if (!templateId) throw new Error("template_id is required");
+
+  if (org && project) {
+    return `/v1/orgs/${encodeURIComponent(org)}/projects/${encodeURIComponent(project)}/templates/${encodeURIComponent(templateId)}`;
+  }
+  if (org) {
+    return `/v1/orgs/${encodeURIComponent(org)}/templates/${encodeURIComponent(templateId)}`;
+  }
+  return `/v1/templates/${encodeURIComponent(templateId)}`;
+}
+
+/**
+ * Builds the NG delete path for templates.
+ * When version_label is provided: /template/api/templates/{id}/{version}
+ * When omitted: /template/api/templates/{id} (deletes all versions)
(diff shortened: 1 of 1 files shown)


hingerabhinav (approved):

hingerabhinav (approved):

prachishah-harness (approved):

scripts/sync-schemas.js
```diff
@@ -124,6 +124,7 @@ ${props}
 
 export const VALID_SCHEMAS = Object.keys(SCHEMAS) as AllSchemaKeys[];
 export type SchemaName = AllSchemaKeys;
+export type SchemaEntry = Record<string, any>;
```
yogesh-chauhan: not needed now

src/data/schemas/index.ts
```diff
@@ -33,6 +33,7 @@ export const SCHEMAS: Record<AllSchemaKeys, Record<string, any>> = {
 
 export const VALID_SCHEMAS = Object.keys(SCHEMAS) as AllSchemaKeys[];
 export type SchemaName = AllSchemaKeys;
+export type SchemaEntry = Record<string, any>;
```
yogesh-chauhan: not needed anymore. You can re-base

yogesh-chauhan (changes requested):

src/registry/toolsets/templates.ts
```diff
@@ -137,17 +184,24 @@ export const templatesToolset: ToolsetDefinition = {
...
-          pathParams: { template_id: "templateIdentifier" },
-          queryParams: { version_label: "versionLabel" },
+          pathBuilder: (input) => {
+            const templateId = input.template_id as string;
+            if (!templateId) throw new Error("template_id is required");
+            const version = input.version_label as string | undefined;
+            if (version) {
+              return `/template/api/templates/${encodeURIComponent(templateId)}/${encodeURIComponent(version)}`;
+            }
+            return `/template/api/templates/${encodeURIComponent(templateId)}`;
```
yogesh-chauhan: can we use a method here as well like `templateV1BasePath` ?

yogesh-chauhan (changes requested):

yogesh-chauhan (approved):

PR #207, opened 2026-05-15
fix: warn on invalid auto-approve-risk header; make config required in write tools

sunilgattupalle:
## Summary

Follow-up to #204 — two review nits that were in a commit on the PR branch but didn't make it into the merge.

- **`session-headers.ts`**: log a `warn` when `X-Harness-Auto-Approve-Risk` contains an unrecognized value (e.g. `"read"`, typos) instead of silently falling back to the deployment default. Prevents confusing silent behavior where a misconfigured client gets a different security posture than intended.
- **Write tool registrars** (`harness_create`, `harness_update`, `harness_delete`, `harness_execute`): change `config?: Config` → `config: Config`. The parameter is always passed by `registerAllTools` — making it optional was a type lie that would let future callers omit it and silently use the process-global auto-approve default instead of the session value.

## Test plan
- [ ] `pnpm typecheck` passes
- [ ] `pnpm test` passes
- [ ] Sending an invalid `X-Harness-Auto-Approve-Risk: read` header in an HTTP session logs a warning and uses the deployment default

### DIFF
--- a/src/tools/harness-create.ts
+++ b/src/tools/harness-create.ts
@@ -10,7 +10,7 @@ import { applyUrlDefaults } from "../utils/url-parser.js";
 import { coerceRecord } from "../utils/type-guards.js";
 import { resourceTypeSchema } from "./input-schemas.js";
 
-export function registerCreateTool(server: McpServer, registry: Registry, client: HarnessClient, config?: Config): void {
+export function registerCreateTool(server: McpServer, registry: Registry, client: HarnessClient, config: Config): void {
   const creatableTypes = registry.getTypesForOperation("create");
 
   server.registerTool(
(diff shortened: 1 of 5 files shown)


thisrohangupta:
Both fixes here are still valid and needed on `main` — verified: neither the invalid-header warning nor the `config?: Config` → `config: Config` tightening exists on current `main`, and the `config?.` guards do gate `HARNESS_READ_ONLY` / auto-approve, so the "type lie" is a real foot-gun.

Since this branch is conflicting + CI-failing + CLA-unblocked, I re-implemented the same two changes cleanly on a fresh branch off `main` and fixed the stale `tool-handlers.test.ts` (plus the other test files that registered write tools without `config`), in #566. Full suite green there (2513 tests).

Suggest closing this as superseded by #566 — leaving that to you since it's your PR.

thisrohangupta:
Closing as superseded by #566, which lands the same two fixes on a clean branch: warn on an unrecognized `X-Harness-Auto-Approve-Risk` header value in `session-headers.ts`, and make `config` required in the write tools. This branch is now conflicting against `main`. Thanks for the original catch — the change is carried forward in #566.

PR #189, opened 2026-05-13
fix: fix log blob download auth for internal MCP mode

Saranya-jena:
## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Documentation
- [ ] Other

## Checklist
- [x] Tests pass
- [x] Typecheck passes

PR #214, opened 2026-05-16
fix: rewrite *.harness.io blob hostname for log downloads on self-managed deployments

mehulparmariitr:
## Description

Fixes log download failures introduced by PR #189 and incompletely addressed by PR #195.

### Root cause chain

**PR #189** (Saranya Jena) added `isPresignedUrl()` and routed any URL carrying `X-Amz-Signature` / `X-Goog-Signature` to `fetch()` directly. The Harness log-service always returns blob links on `app.harness.io/storage/...` regardless of `HARNESS_BASE_URL`, so self-managed deployments where `app.harness.io` is unreachable started getting `TypeError: fetch failed`.

**PR #195** attempted to fix this by routing `*.harness.io` pre-signed URLs through `client.requestStream()`. This stopped the network error but introduced HTTP 403: the request became `{HARNESS_BASE_URL_HOST}/gateway/storage/harness-download/...` — the API gateway does not serve `/storage/...` CDN paths.

**Verified on a self-managed deployment:**
- `{gateway-host}/gateway/storage/harness-download/...` → **HTTP 403** ❌
- `{gateway-host}/storage/harness-download/...` → **HTTP 200** ✅

### Fix

Three routing strategies in `downloadBlobContent()`:

| URL type | Strategy | Why |
|---|---|---|
| S3 / GCS hostnames | Direct fetch — original URL unchanged | AWS/GCS signatures are host-bound; rewriting invalidates them |
| `*.harness.io` + pre-signed + `/storage/` path | Rewrite hostname+port to `HARNESS_BASE_URL` host → direct fetch | CDN path; gateway 403s; pre-signed params authenticate the request |
| All other `*.harness.io` | `client.requestStream()` + `/gateway/log-service` prefix | API path — needs auth header injection |

The `/storage/` path guard on Strategy 2 prevents misclassification: a `*.harness.io` log-service API path (e.g. `/gateway/log-service/...`) that happens to carry pre-signed params would otherwise be direct-fetched without auth headers, causing 401/403.

Additional hardening:
- **Port preserved**: `blobUrl.port = baseUrl.port` — prevents silent port drop for non-default-port deployments
- **Unparseable baseURL throws**: instead of falling back to direct-fetching the blocked `app.harness.io` URL

### Tests (22 total, 9 strict invariants)

| Test | What it catches |
|---|---|
| `requestStream` NOT called for CDN blobs | PR #195 regression: routing through gateway |
| fetch URL does NOT contain `/gateway/storage` | The exact path that produces HTTP 403 |
| Pre-signed query params preserved after host rewrite | Dropped `X-Amz-Signature` → CDN 403 |
| GCS pre-signed on Harness CDN also host-rewritten | X-Goog-Signature case was missing from prior suites |
| S3/GCS external hosts NOT rewritten | Rewriting breaks the AWS/GCS signature |
| Non-presigned Harness URLs still go through `requestStream` | Auth injection required for API paths |
| baseURL non-default port preserved in rewritten URL | Silent port drop routes to wrong server |
| Unparseable baseURL throws, does NOT fetch `app.harness.io` | Misconfigured base URL should fail clearly |
| Presigned `*.harness.io` on non-`/storage/` path → `requestStream` | Prevents Strategy 2 from misrouting signed API paths |

## Type of Change
- [x] Bug fix

## Checklist
- [x] Tests pass (1357 total)
- [x] Typecheck passes

### DIFF
--- a/src/utils/log-resolver.ts
+++ b/src/utils/log-resolver.ts
@@ -368,8 +368,10 @@ async function downloadBlobContent(
     }
   }
 
-  if (blobUrl && isPresignedUrl(blobUrl) && isHarnessHost(blobUrl.hostname)) {
+  if (blobUrl && isPresignedUrl(blobUrl) && isHarnessHost(blobUrl.hostname) && blobUrl.pathname.startsWith("/storage/")) {
     // Strategy 2: *.harness.io CDN blob — rewrite hostname and direct fetch.
+    // Guarded by /storage/ prefix to avoid misrouting signed API paths (e.g. /gateway/log-service/...)
+    // that happen to carry pre-signed params — those must go through client.requestStream() for auth.
     // The blob link always points to app.harness.io/storage/... regardless of HARNESS_BASE_URL.
     // Rewrite to the configured host so self-managed deployments can reach it.
     const baseUrl = safeParseUrl(client.baseURL);
--- a/src/utils/log-resolver.ts
+++ b/src/utils/log-resolver.ts
@@ -339,17 +339,23 @@ async function downloadBlobContent(
 ): Promise<Response> {
   const blobUrl = safeParseUrl(blobLink);
 
-  // Two routing strategies based on the blob URL:
+  // Three routing strategies based on the blob URL:
   //
   // 1. True external storage (S3, GCS domains) → direct fetch, no auth headers.
   //    The URL is publicly routable with embedded signature params; routing through
   //    the client or adding extra headers would invalidate the AWS/GCS signature.
   //
-  // 2. All other URLs (including *.harness.io pre-signed) → client.requestStream()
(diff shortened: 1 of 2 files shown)


sunilgattupalle (approved):
Thanks @mehulparmariitr — for the root cause analysis across three PRs is clearly documented, the three-strategy routing table is the right model (signature-bound external hosts, Harness CDN host-rewrite, API paths via client), and the /storage/ path guard is the key correctness invariant that prevents Strategy 2 from misrouting signed API paths.

One minor nit not worth holding up the PR: the function-level docstring on downloadBlobContent still says "Two routing strategies" and "Harness-hosted URLs are routed through the client" — it wasn't updated to reflect the three-strategy model. Fine to fix in a follow-up.

thisrohangupta (approved):

PR #230, opened 2026-05-19
fix: sync ai-evals toolset with current API state

SrikarMannepalli:
## Summary
- Fix target type enums (`prompt|agent|precomputed`, not `app|static`) and metric type description
- Add git-backed storage fields (`storage_type`, `git_source`) to all entity CRUD schemas
- Add missing fields: `connector_ref`, `schedule`, `branch`, `judge_model_id`, `expected_tool_calls`, `git_commit_sha`, `config` on metric set entries
- Add `search` query param to dataset, eval, target, model, metric-set list operations
- Add `target_id` filter to eval list and runs list
- Add `export_yaml` and `overview` execute actions to `eval_target`
- Add `bulk_upsert` execute action for dataset items
- Remove phantom `eval_id`/`status` filters from runs list (not supported by API)
- Update resource descriptions and prompts for git-backed storage and scheduling

## Test plan
- [x] `pnpm typecheck` passes
- [x] Verify `harness_describe(resource_type="eval_target")` shows new execute actions
- [x] Verify `harness_describe(resource_type="evaluation")` shows updated list filters
- [x] Spot-check `harness_schema(resource_type="eval_dataset")` for `storage_type`/`git_source` fields

### DIFF
--- a/src/prompts/create-eval-suite.ts
+++ b/src/prompts/create-eval-suite.ts
@@ -33,7 +33,7 @@ export function registerCreateEvalSuitePrompt(server: McpServer): void {
 
 1. **List existing evaluations** — \`harness_list(resource_type="evaluation")\` — pick eval IDs to include.
 
-2. **Create the suite** — \`harness_create(resource_type="eval_suite", body={ name: "${suite_name}", pass_strategy: "all_must_pass" | "weighted_threshold", pass_threshold?: number })\`
+2. **Create the suite** — \`harness_create(resource_type="eval_suite", body={ name: "${suite_name}", pass_strategy: "all_must_pass" | "weighted_threshold", pass_threshold?: number, schedule?: { cron: "0 8 * * *", timezone?: "UTC", enabled?: true } })\`
 
 3. **Add members** (choose one pattern)
    - **Add one at a time**: \`harness_create(resource_type="eval_suite_evaluation", params={ suite_id: "<SUITE_ID>" }, body={ evaluation_id: "<EVAL_ID>", is_required: true })\`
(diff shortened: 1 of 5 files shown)


sunilgattupalle (commented):
## Code Review

**Overall: Approve with minor suggestions** — well-scoped schema sync, low risk, good PR description.

---

### 1. Breaking change: `createTargetSchema` relaxes required fields without guard

`type` and `config` went from `required: true` to `required: false`. The per-field description says "required when storage_type='managed'" but nothing enforces or surfaces this at the schema level. An LLM agent could omit both for a managed target and get a confusing backend error.

**Suggestion:** Update `createTargetSchema.description` to make the conditional requirement explicit:
```typescript
description: "Create target. For managed: type + config are required. For git-backed: omit type/config, provide storage_type + git_source.",
```

### 2. `bulk_upsert` uses PATCH — intentional?

```typescript
bulk_upsert: { method: "PATCH", ... }
```
Upsert semantics via PATCH is uncommon (most bulk upsert endpoints use POST or PUT). If the backend genuinely expects PATCH here, fine — just flagging for a double-check.

### 3. Runs list lost `eval_id`/`status` filters — migration note

The list description now says "Use eval_run_by_eval resource to filter by eval_id" which is good. Consider also adding this hint to the `eval_run` resource-level description so LLM agents discover the migration path without reading operation descriptions.

### 4. Minor: `item_identifier` vs `item_id` naming

In `testTargetSchema`, the new field is `item_identifier` while the rest of the file uses `_id` suffix patterns. If the backend expects this exact name, keep it — just noting the inconsistency.

---

Nice cleanup overall. The git-backed storage fields are consistent across entities and the removed phantom filters are good hygiene.

SrikarMannepalli:
1. createTargetSchema — Updated description to explicitly state the conditional: "For managed: type + config are required. For git-backed: omit type/config..."
2. bulk_upsert PATCH — Verified that backend does use patch
3. eval_run description — Added hint: "To filter by eval_id, use the eval_run_by_eval resource instead."
4. item_identifier naming — Verified that the backend does use item_identifier field

sunilgattupalle (approved):

PR #242, opened 2026-05-21
fix(iacm): fix iacm_resource output validation and iacm_module pagination

### LINKED ISSUE #240: iacm_resource list fails with output validation error on workspaces without state
## Description

When calling `harness_list(resource_type='iacm_resource', filters={workspace_id: '...'})` on a workspace that has never been applied (no Terraform state), the MCP server returns an output validation error instead of an empty resource list.

## Error

```
MCP error -32602: Output validation error: Invalid structured content for tool harness_list:
[
  {
    "expected": "array",
    "code": "invalid_type",
    "path": ["items"],
    "message": "Invalid input: expected array, received undefined"
  }
]
```

## Root Cause

The `iacmResourcesExtract` function in `src/registry/toolsets/iacm.ts` returns a response with a `resources` field, but the MCP framework's output schema for `harness_list` expects an `items` field. When the workspace has no state, the API likely returns `null` or an empty structure that doesn't map to the expected output shape.

The extractor returns:
```ts
{
  resources,        // ← framework expects this to be called "items"
  outputs,
  data_sources,
  page_count,
  ...
}
```

## Steps to Reproduce

1. Create an IaCM workspace (never run apply)
2. Call `harness_list(resource_type='iacm_resource', filters={workspace_id: '<workspace_id>'})`
3. Observe output validation error

## Expected Behavior

Should return an empty list response like:
```json
{"items": [], "page_count": 0, "has_more": false}
```

## Environment

- harness-mcp-v2@3.0.5
- HARNESS_TOOLSETS=+iacm

### LINKED ISSUE #241: iacm_module list fails with page=0 default despite pageOneIndexed: true
## Description

Calling `harness_list(resource_type='iacm_module')` without an explicit page filter fails because the MCP framework sends `page=0` to the IaCM API, which requires 1-based pagination.

## Error

```
"page must be greater or equal than 1 but got value 0"
```

## Root Cause

In `src/registry/toolsets/iacm.ts`, the `iacm_module` list operation correctly sets `pageOneIndexed: true`, but the framework doesn't appear to convert the default page value (0) to 1 when constructing the API request. The conversion only seems to apply when an explicit page parameter is passed by the caller.

```ts
operations: {
  list: {
    // ...
    pageOneIndexed: true,  // ← set correctly, but default page=0 still sent
  }
}
```

## Steps to Reproduce

1. Set `HARNESS_TOOLSETS=+iacm`
2. Call `harness_list(resource_type='iacm_module')` (no page filter)
3. Observe error: "page must be greater or equal than 1 but got value 0"

## Workaround

Explicitly pass `filters: {page: 1}`:
```
harness_list(resource_type='iacm_module', filters={page: 1})
```

## Expected Behavior

When `pageOneIndexed: true` is set and no page is specified, the framework should default to page=1 (not 0).

## Environment

- harness-mcp-v2@3.0.5
- HARNESS_TOOLSETS=+iacm

dtsong-harness:
## Description

Fixes two related bugs in the IaCM toolset (introduced in v3.0.5, PR #233):

1. **`iacm_resource` list fails with output validation error** (#240): The `iacmResourcesExtract` function returned a `resources` field, but the `listOutputSchema` requires `items`. Renamed the field to `items` so structured output validation passes. The `outputs` and `data_sources` fields remain as additional context alongside the standard `items` array.

2. **`iacm_module` list fails with "page must be >= 1"** (#241): The `iacm_module` list operation was missing `pageOneIndexed: true`, causing the framework's default `page=0` to be sent directly to the IaCM API which requires 1-based pagination.

### Reproduction

Both bugs verified against a live Harness account (`harness-mcp-v2@3.0.5` with `HARNESS_TOOLSETS=+iacm`):
- `iacm_resource`: Call `harness_list` on a workspace with no Terraform state → output validation error
- `iacm_module`: Call `harness_list` without explicit page filter → "page must be >= 1"

### Changes

- `src/registry/toolsets/iacm.ts`: Rename `resources` → `items` in `iacmResourcesExtract`; add `pageOneIndexed: true` to `iacm_module` list
- `tests/registry/iacm.test.ts`: Update assertion to check `items` instead of `resources`

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Documentation
- [ ] Other

## Checklist
- [x] Tests pass (`pnpm test` — 1432 tests, 60 files)
- [x] Typecheck passes (`pnpm build` clean)

Fixes #240, Fixes #241

### DIFF
--- a/src/registry/toolsets/iacm.ts
+++ b/src/registry/toolsets/iacm.ts
@@ -52,7 +52,7 @@ const iacmResourcesExtract = (raw: unknown): unknown => {
   const has_more = r.hasMore ?? false;
   const page_count = resources.length;
   return {
-    resources,
+    items: resources,
     outputs: r.outputs ?? [],
     data_sources: r.data_sources ?? [],
     page_count,
@@ -389,6 +389,7 @@ export const iacmToolset: ToolsetDefinition = {
             page: "page",
             size: "size",
           },
+          pageOneIndexed: true,
           operationPolicy: { risk: "read", retryPolicy: "safe" },
           responseExtractor: moduleListExtract,
           description:
--- a/tests/registry/iacm.test.ts
+++ b/tests/registry/iacm.test.ts
@@ -226,7 +226,7 @@ describe("iacmResourcesExtract", () => {
     return getOp("iacm_resource", "list").responseExtractor!(raw) as Record<string, unknown>;
   }
 
-  it("extracts resources, outputs, data_sources sections", () => {
+  it("extracts resources as items, outputs, data_sources sections", () => {
     const raw = {
       resources: [{ name: "aws_instance.main" }],
       outputs: [{ name: "vpc_id", value: "vpc-123" }],
@@ -235,7 +235,7 @@ describe("iacmResourcesExtract", () => {
       totalItems: 1,
     };
     const result = extract(raw);
-    expect((result.resources as unknown[]).length).toBe(1);
+    expect((result.items as unknown[]).length).toBe(1);
     expect((result.outputs as unknown[]).length).toBe(1);
     expect(result.total_items).toBe(1);
     expect(result.has_more).toBe(false);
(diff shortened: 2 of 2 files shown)


thisrohangupta:
this was fixed by engineering, just merged engineers fix

PR #301, opened 2026-06-04
fix: document pull request repo params

skjindal93:
## Summary
- Add `paramsSchema` metadata so `harness_describe` exposes required pull request params like `repo_id` and `pr_number`.
- Document PR-related repo params across pull request resources and execute actions.
- Accept `repo_identifier` as a compatibility alias for `repo_id` when resolving repo path params.

## Test plan
- `pnpm test -- tests/registry/pull-requests.test.ts tests/tools/tool-handlers.test.ts`
- `pnpm typecheck`
- Local MCP smoke test against harness0 confirmed `repo_identifier` reaches the Harness API instead of failing MCP validation.

### DIFF
--- a/src/registry/index.ts
+++ b/src/registry/index.ts
@@ -540,6 +540,13 @@ export class Registry {
         for (const [inputKey, pathPlaceholder] of Object.entries(spec.pathParams)) {
           let value = input[inputKey];
           if (value === undefined || value === "") {
+            // Accept repo_identifier as an alias for repo_id — agents sometimes use
+            // the GitOps/Repositories naming convention instead of the Code PR convention.
+            if (inputKey === "repo_id" && input["repo_identifier"] !== undefined && input["repo_identifier"] !== "") {
+              value = input["repo_identifier"];
+            }
+          }
+          if (value === undefined || value === "") {
             // Default scope placeholders from config for project/org-scoped resources
             if (pathPlaceholder === "org" && shouldUseOrg(pathDefaultScope)) {
               value = explicitScopeValues?.orgId ?? this.config.HARNESS_ORG;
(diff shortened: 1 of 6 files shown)


src/registry/index.ts
```diff
@@ -539,6 +539,13 @@ export class Registry {
       if (spec.pathParams) {
         for (const [inputKey, pathPlaceholder] of Object.entries(spec.pathParams)) {
           let value = input[inputKey];
+          if (value === undefined || value === "") {
+            // Accept repo_identifier as an alias for repo_id — agents sometimes use
+            // the GitOps/Repositories naming convention instead of the Code PR convention.
+            if (inputKey === "repo_id" && input["repo_identifier"] !== undefined && input["repo_identifier"] !== "") {
```
sunilgattupalle: Is this needed if we are declaring param schema?

sunilgattupalle (approved):

src/tools/harness-describe.ts
```diff
@@ -52,13 +52,15 @@ export function registerDescribeTool(server: McpServer, registry: Registry): voi
               operation: op,
               method: spec.method,
               description: spec.description,
```
sunilgattupalle, line 54: The `?? undefined` here is a no-op — `spec.paramsSchema` is already `undefined` when absent, so the nullish coalesce changes nothing. Just use `spec.paramsSchema` directly (same as the pre-existing `bodySchema` lines above this diff, which have the same issue but weren't touched here).

sunilgattupalle:
**PR description mismatch:** The summary says *"Accept `repo_identifier` as a compatibility alias for `repo_id`"* but the diff does the opposite — it adds a test asserting `repo_identifier` is rejected with an error. The alias feature either didn't make it in or was intentionally dropped. Either way, please update the description before merge so the git history isn't misleading.

PR #320, opened 2026-06-10
fix: preserve caller-provided FME auth

Saranya-jena:
## Summary
- Move FME default authentication from the registry layer into `HarnessClient`, so caller-provided `Authorization` can be preserved before fallback auth is applied.
- Avoid sending placeholder API keys such as `dummy` or `pat.*.*.dummy` to Split/FME.
- Keep standalone behavior working by falling back to a non-placeholder request `x-api-key`, `HARNESS_FME_API_KEY`, or an FME-entitled `HARNESS_API_KEY` for FME calls.
- Leave non-FME Harness toolsets on the existing `x-api-key` default auth path.

## Why
FME/Split Admin APIs use Bearer auth, while the rest of the Harness APIs commonly use `x-api-key`. Keeping FME auth at the registry layer made it harder for callers to supply their own `Authorization` header. Moving fallback auth into the client makes the behavior consistent: explicit auth wins, FME fallback credentials are normalized to Bearer, and placeholder credentials are blocked before any request is sent to Split/FME.

## Behavior Covered
- FME requests with caller-provided `Authorization` are passed through unchanged.
- FME requests with a valid non-placeholder API key are sent as `Authorization: Bearer <key>`.
- FME requests with only placeholder credentials fail before making a network call.
- Regular Harness requests continue to receive `x-api-key` when no auth header is supplied.

## Test Plan
- [x] `pnpm test tests/client/harness-client.test.ts tests/registry/feature-flags.test.ts`
- [x] Local smoke test with QA FME token against `https://api.split-stage.io/internal/api/v2/workspaces` returned 200.
- [x] Local MCP `harness_list` for `fme_workspace` returned workspace data through the updated config.

### DIFF
--- a/src/client/harness-client.ts
+++ b/src/client/harness-client.ts
@@ -155,9 +155,13 @@ export class HarnessClient {
   }
 
   private applyDefaultAuth(headers: Record<string, string>, isFme: boolean): void {
+    // Hosted/internal MCP wrappers may already provide service-to-service auth.
+    // Preserve it instead of layering the configured PAT/SAT on top.
     if (headers["Authorization"]) return;
 
     if (isFme) {
+      // FME/Split Admin APIs expect Bearer auth. Drop x-api-key here so hosted
+      // placeholder credentials are never forwarded to api.split.io.
       const headerApiKey = headers["x-api-key"]?.trim();
       delete headers["x-api-key"];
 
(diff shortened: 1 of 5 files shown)


Saranya-jena:
@cursoragent can you review the PR

prachishah-harness (approved):

thisrohangupta:
@cursoragent can you review this PR?

thisrohangupta (approved):

PR #359, opened 2026-06-17
Incident and deploy types

piis3:
## Description
Adds support for AI-SRE incident types and AI-SRE detected deployment types.

## Type of Change
- [ ] Bug fix
- [x] New feature
- [ ] Refactor
- [ ] Documentation
- [ ] Other

## Checklist
- [x] Tests pass
- [x] Typecheck passes


thisrohangupta:
@cursoragent can you run Sunil's Architecture Review on this?

src/registry/toolsets/incidents.ts
```diff
@@ -0,0 +1,144 @@
...
+          responseExtractor: offsetListExtract,
+          description: "List incidents with filtering by status, severity, service, environment, commander, and text",
+        },
+        get: {
+          method: "GET",
+          path: "/gateway/ir/tp/api/v1/mc/incidents/{incidentId}",
+          operationPolicy: { risk: "read", retryPolicy: "safe" },
+          pathParams: { incident_id: "incidentId" },
+          responseExtractor: passthrough,
+          description: "Get incident details by ID",
```
sunilgattupalle, line 193: **Architecture violation: `passthrough` on real endpoints**

All four incident operations (get, create, update, close) use `responseExtractor: passthrough`, which leaks raw backend envelope/debug/meta fields across the tool boundary.

CLAUDE.md explicitly prohibits this: *"Never `passthrough` on a real endpoint. Backend envelope/debug/meta must not cross the tool boundary."*

The `deploys.ts` in this same PR demonstrates the correct pattern with `deployGetExtract`. Incidents need an equivalent `incidentGetExtract` that projects a stable, documented shape (`prettyId`, `title`, `status`, `severity`, `summary`, `impactedServices`, `environments`, `commanderHarnessUserId`, `createdAt`, `updatedAt`).

src/registry/toolsets/deploys.ts
```diff
@@ -0,0 +1,126 @@
...
+    slim.summary = item.summary.split("\n", 1)[0];
+  }
+  if (Array.isArray(item.environments)) slim.environments = item.environments;
+  if (Array.isArray(item.buildVersions)) {
+    slim.services = item.buildVersions.map((bv) => {
+      const b = (bv ?? {}) as Record<string, unknown>;
+      return { service: b.service, version: b.version };
+    });
+  }
+  if (item.deployTimestamp !== undefined) slim.deployTimestamp = item.deployTimestamp;
```
sunilgattupalle: **Inconsistent null handling for `deployTimestamp` between list and get**

`compactDeploy` uses `if (item.deployTimestamp !== undefined)` which keeps `null`, but `deployGetExtract` uses `if (typeof raw.deployTimestamp === 'number')` which drops `null`.

An agent that lists deploys will see `deployTimestamp: null` and then fetch the detail view expecting the same field, only to find it absent. Recommend aligning both to `typeof === 'number'` since a real deploy timestamp is never null.

src/registry/toolsets/deploys.ts
```diff
@@ -0,0 +1,126 @@
...
+function compactDeploy(item: Record<string, unknown>): Record<string, unknown> {
+  const slim: Record<string, unknown> = {};
+  if (typeof item.id === "string") slim.id = item.id;
+  if (typeof item.projectId === "string") slim.projectId = item.projectId;
+  if (typeof item.summary === "string" && item.summary.length > 0) {
+    slim.summary = item.summary.split("\n", 1)[0];
+  }
+  if (Array.isArray(item.environments)) slim.environments = item.environments;
+  if (Array.isArray(item.buildVersions)) {
+    slim.services = item.buildVersions.map((bv) => {
```
sunilgattupalle, line 25: **`compactDeploy` lacks `isRecord` guard on `buildVersions` items**

`deployGetExtract` (same file) correctly uses `if (!isRecord(bv)) return bv` before accessing `bv.service` / `bv.version`. `compactDeploy` instead uses `(bv ?? {})` which coerces `null` to `{}` but leaves strings/numbers as-is, then casts to `Record<string, unknown>`.

If the API ever returns a non-object element in `buildVersions`, the result is `{ service: undefined, version: undefined }` silently emitted into the list output. The `isRecord` guard from `type-guards.ts` is already imported in this file — use it here the same way `deployGetExtract` does.

src/tools/harness-list.ts
```diff
@@ -79,7 +79,8 @@ export function registerListTool(server: McpServer, registry: Registry, client:
         if (args.compact !== false && !resultSkipCompact && isRecord(result)) {
           const items = result.items;
           if (Array.isArray(items)) {
-            result.items = compactItems(items);
+            const compactFn = registry.getResource(resourceType).compactItem;
+            result.items = compactItems(items, compactFn);
           }
         }
```
sunilgattupalle, line 86: **`harness-search` doesn't receive `compactFn` — per-resource compaction silently skipped for search results**

`harness-list.ts` is updated to pass `registry.getResource(resourceType).compactItem` into `compactItems`. But `harness-search.ts` still calls `compactItems(r.items)` with the old single-arg signature.

This means searching for `deploy` resources returns the generic whitelist-compacted shape (missing `services`, untruncated `summary`), while listing returns the correct `compactDeploy` shape. An agent doing search + list for the same resource type will see two different response shapes for the same data.

src/registry/toolsets/deploys.ts
```diff
@@ -0,0 +1,126 @@
...
+import { isRecord } from "../../utils/type-guards.js";
+
+/**
+ * Same custom scope naming as the incidents API (both route through
+ * /api/v1/mc/ behind the @HarnessAuth filter): accountId / orgId / projectId
+ * rather than the standard NG accountIdentifier / orgIdentifier /
+ * projectIdentifier. The client still appends its default `accountIdentifier`
+ * query param; the Java side ignores unknown params.
+ */
+const DEPLOY_SCOPE = { account: "accountId", org: "orgId", project: "projectId" } as const;
```
sunilgattupalle: **`DEPLOY_SCOPE` and `INCIDENT_SCOPE` are identical — and duplicate `STO_SCOPE`**

Both are `{ account: 'accountId', org: 'orgId', project: 'projectId' } as const`, and this is already the third definition of this shape (STO has the same in `sto.ts`). The comment in both files even says "both route through /api/v1/mc/".

This should be a single shared constant (e.g. `MC_SCOPE`) exported from a shared location. If the Harness platform ever renames one of these query params (it has happened between API generations), only one of the three copies will get fixed.

src/utils/compact.ts
```diff
@@ -59,10 +59,18 @@ function isWhitelistedKey(key: string): boolean {
...
+ *
+ * When `compactFn` is provided (from a resource's `compactItem`), it replaces
+ * the generic key-name whitelist for object items — used by resources whose
+ * useful fields can't be expressed as a flat whitelist (e.g. deploys).
  */
-export function compactItems(items: unknown[]): unknown[] {
+export function compactItems(
+  items: unknown[],
+  compactFn?: (item: Record<string, unknown>) => Record<string, unknown>,
+): unknown[] {
```
sunilgattupalle, line 71: **`compactFn` path bypasses the `openInHarness` → markdown hyperlink transform**

When `compactFn` is provided, `compactItems` returns early from the custom function and never reaches the post-loop block that merges `openInHarness` into `name` as a markdown link (the ```slim.name = `[...](url)``` logic).

If a deploy item ever gains an `openInHarness` field, the link will either leak as a raw URL (if `compactDeploy` preserves it) or be silently dropped — neither is the intended behavior. The `compactFn` contract should document this gap, or the transform should be applied after `compactFn` returns.

thisrohangupta (approved):

PR #363, opened 2026-06-19
fix: resolve nested harness_schema paths like EnvironmentV1

Saranya-jena:
Add recursive schema path navigation so harness_schema can find nested type definitions (e.g. pipeline_v1/stages/unified/EnvironmentV1) when agents request a single-segment path.

## Type of Change
- [ ] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Documentation
- [ ] Other

## Checklist
- [ ] Tests pass
- [ ] Typecheck passes


thisrohangupta:
Closing: superseded. Nested schema-path resolution already landed on `main` — `harness-schema.ts` has `findDefinitionByName` with resolution order "direct key → literal dot-path → recursive name search" and resolves bare nested names like `EnvironmentV1`. The separate `schema-path-navigation.ts` module is no longer needed. Verified against current main.

Thanks for the fix — the capability it added is live on main via a different implementation.

PR #444, opened 2026-06-25
feat: default KG queryable-type discovery to a configurable annotation cohort

nandesh-harness:
## Summary

`kg_queryable_type_summary` discovery (`getQueryableTypes`) only filters by annotation when the caller passes `annotations` on the request. With no way to set a server-side default, scoping discovery to a curated cohort depends on every caller remembering to pass the right annotation — and agents often don't, so they discover the full unfiltered catalog.

This adds a `KG_DEFAULT_ANNOTATIONS` env var (comma-separated) that `queryableTypesBody` falls back to **only when the caller supplies no annotations**.

## Scope

- `src/registry/toolsets/knowledge-graph.ts` — `DEFAULT_KG_ANNOTATIONS` parsed from `process.env.KG_DEFAULT_ANNOTATIONS`; `queryableTypesBody` uses it as the fallback.
- `tests/registry/knowledge-graph.test.ts` — 4 cases covering the filter-body construction.

## Behavior

| caller `annotations` | `KG_DEFAULT_ANNOTATIONS` | filter sent |
|---|---|---|
| supplied | (any) | caller's values (unchanged) |
| none | unset / empty | no annotations filter (unchanged) |
| none | set | the configured default |

Backward-compatible: unset/empty preserves today's unfiltered behavior, and an explicit caller value always takes precedence over the default.

## Why this is safe

- Pure-data toolset change: logic stays inside `queryableTypesBody`; no new client/HTTP/stdout. Default is empty, so existing deployments behave identically until they opt in.

## Test plan

- [x] `pnpm run typecheck` — clean
- [x] `pnpm exec vitest run tests/registry/knowledge-graph.test.ts` — 21/21 pass (4 new)

## Rollback

Revert the commit; with `KG_DEFAULT_ANNOTATIONS` unset there is no behavioral change to undo.

### DIFF
--- a/src/registry/toolsets/knowledge-graph.ts
+++ b/src/registry/toolsets/knowledge-graph.ts
@@ -3,6 +3,15 @@ import type { ToolsetDefinition, FilterFieldSpec } from "../types.js";
 const QUERY_SVC =
   "/query-service/grpc/io.harness.platform.query.service.api.v1.QueryServiceGrpc";
 
+// Default annotation cohort applied to queryable-type discovery when the caller
+// supplies none. Empty (unset) preserves the prior behaviour of returning the
+// full unfiltered catalog; set KG_DEFAULT_ANNOTATIONS (comma-separated) per
+// deployment to scope discovery to a cohort.
+const DEFAULT_KG_ANNOTATIONS = (process.env.KG_DEFAULT_ANNOTATIONS ?? "")
+  .split(",")
+  .map((a) => a.trim())
+  .filter(Boolean);
+
 // ─── Response extractors ─────────────────────────────────────────────────────
 
 const MAX_DESC_LEN = 80;
@@ -122,7 +131,9 @@ function queryableTypesBody(input: Record<string, unknown>): Record<string, unkn
     filter.kinds = Array.isArray(kinds) ? kinds : [kinds];
   }
 
-  const annotations = input.annotations;
+  const annotations =
+    input.annotations ??
+    (DEFAULT_KG_ANNOTATIONS.length > 0 ? DEFAULT_KG_ANNOTATIONS : undefined);
   if (annotations) {
     filter.annotations = Array.isArray(annotations) ? annotations : [annotations];
   }
--- a/tests/registry/knowledge-graph.test.ts
+++ b/tests/registry/knowledge-graph.test.ts
@@ -220,6 +220,78 @@ describe("kg_queryable_type_summary list extractor", () => {
 });
 
 // ---------------------------------------------------------------------------
+// kg_queryable_type_summary — request-body filter construction
+// ---------------------------------------------------------------------------
+
+describe("kg_queryable_type_summary list body filter", () => {
+  // KG_DEFAULT_ANNOTATIONS is read at module load, so each env permutation
+  // re-imports the registry with a fresh module graph.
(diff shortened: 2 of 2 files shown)


thisrohangupta:
Verdict: Approve with minor nits — the change is small, backward-compatible, and well-tested for the core contract. A few documentation and convention gaps… Every other deployment knob (HARNESS_TOOLSETS, HARNESS_PIPELINE_VERSION, HARNESS_DIAGNOSE_LOG_FETCH_CONCURRENCY, etc.) goes through ConfigSchema in src/config.ts. This PR reads process.env directly in the toolset module.

That works, but it breaks the established pattern:
• Not discoverable via makeConfig() in tests (hence the resetModules workaround)
• Won't appear in config validation tests
• Can't be surfaced in startup logs the way other config is

Suggestion: Add KG_DEFAULT_ANNOTATIONS to ConfigSchema (comma-split in transform), pass through Registry config, and read from this.config in the body builder (or close over config at registry init). Not required for correctness, but aligns with repo conventions.

2. Missing operator documentation

KG_DEFAULT_ANNOTATIONS is absent from:
• .env.example
• README env var table

Deployers won't discover it without reading the PR.

3. Agent-facing description gap

The resource description says "Pass 'kinds' to filter" but doesn't mention annotations or that a server default may apply. Per anti-bloat rules this belongs in description / list op description, not global server instructions:

typescript
// Suggested addition to list op description:
"Pass kinds and/or annotations to filter. When annotations are omitted, the server may apply KG_DEFAULT_ANNOTATIONS if configured."


4. Test gaps

Worth adding:
• KG_DEFAULT_ANNOTATIONS=cohort_a,cohort_b → ["cohort_a", "cohort_b"] (validates the parsing logic)
• Default annotations + caller-supplied kinds (both in filter)
• Whitespace: " cohort_a , cohort_b " trimming

---

Worth noting (not blockers)

Module-load-time capture — DEFAULT_KG_ANNOTATIONS is frozen at import. Fine for production (env set at startup); tests already handle via resetModules. Document that it's a startup-time setting.

Empty-array edge case — annotations: [] is explicit (won't fall back to default) and sends filter.annotations: [] because [] is truthy. Pre-existing behavior; only matters if the API treats [] differently from omitted filter.

Multi-user mode — Env-level default applies globally per server instance. That's appropriate for cohort scoping; no per-session override needed.

---

Security / Safety

No concerns. This is a read-only list filter. No secrets exposed, no write-path changes, no confirmation-gating impact.

---

Recommended actions for author

1. Add KG_DEFAULT_ANNOTATIONS to .env.example with a one-line comment.
2. Add a README row in the env var table (Knowledge Graph section or general optional vars).
3. Update the list operation description to mention annotations + optional server default.
4. Add 1–2 tests for comma-separated parsing and kinds+default combo.
5. (Optional but preferred) Route through config.ts instead of raw process.env.

——

This is a sensible, low-risk deployment knob for KG cohort filtering. The body-builder logic and test strategy are sound. Main gaps are documentation (.env.example, README, resource description) and config.ts convention — the kinds of things the architecture review bot typically flags on the first pass.

Recommendation: approve after doc + minor test additions (config.ts refactor optional but would prevent a follow-up nit).

nandesh-harness:
Discarding this PR, since it is not required.

PR #451, opened 2026-06-26
feat: semantic search routing — 98% reduction in scatter-gather API calls

sunilgattupalle:
## Summary

- Adds a pluggable `SearchProvider` interface to mcp-server with `LocalSearchProvider` (ONNX embeddings, `all-MiniLM-L6-v2`, in-memory cosine similarity)
- `harness_search` now runs semantic search **first**, uses high-confidence hits (score ≥ 0.5) to predict which resource types to query, then scatter-gathers only those — reducing API calls from 163 to 1–8 for most queries (~98% reduction)
- Static content indexed at startup into `mcp_resources` corpus (permanent): resource definitions, YAML examples, schemas, entity schemas — all account-agnostic
- Fire-and-forget indexing in `harness_list`/`harness_get` populates `resources` corpus per account as agents use the tools
- Semantic results at tier-0 in search response; response includes `semantic_routed=true` and `types_skipped=N` when routing fires

## How It Works

```
query → LocalSearchProvider.search() → extractRoutingTypes() (score ≥ 0.5)
      → scatter-gather [connector] only  ← was 163 types
      → merge tier-0 semantic hits
```

Resource definitions and examples have `resource_type` in metadata — primary routing signal. Schema entries get comma-separated types for display only (not routing, since schemas span multiple types).

## Test Plan

- [ ] `pnpm build` passes (TypeScript strict, no errors)
- [ ] `pnpm test` passes (vitest env sets `HARNESS_SEARCH_PROVIDER=none` to skip model download in CI)
- [ ] Manual: search "kubernetes connector" → `searched_types: 3`, `types_skipped: 160`, tier-0 mcp_resources results
- [ ] Manual: search "github secret" (ambiguous) → falls back to full scatter-gather, no routing
- [ ] Model cached at `/tmp/hf-cache/Xenova/all-MiniLM-L6-v2/` (~23MB, downloaded once)

## Future Work (tracked in `plans/`)

- Account isolation security review for `resources` corpus
- User-level isolation for user-scoped resource types
- On-demand indexing for dynamic account-agnostic types (kg_queryable_type, global templates)
- `HarnessSearchProvider` — HTTP client to search-service for production cloud deployments
- Docs corpus ingestion
- Routing telemetry (precision/recall measurement)


src/config.ts
```diff
@@ -113,6 +113,11 @@ const RawConfigSchema = z.object({
     emptyStringAsUndefined,
     z.coerce.number().min(1).max(20).default(3),
   ),
+  HARNESS_SEARCH_PROVIDER: z.preprocess(
+    emptyStringAsUndefined,
+    z.enum(["none", "local"]).default("local"),
```
skjindal93, line 118: **[Blocking] Default-on feature with an optional engine.** `HARNESS_SEARCH_PROVIDER` defaults to `local`, but `@huggingface/transformers` is in `optionalDependencies`. If it isn't installed (or no matching native binary for the platform), `LocalSearchProvider.initialize()` swallows the error and the feature silently no-ops. Combined with the runtime model download (see local-provider.ts), the likely production outcome is "works on a laptop, does nothing — or fails to init — in the cluster."

Recommendation: either make `@huggingface/transformers` a regular dependency since `local` is the default, **or** keep it optional and default this to `none` (opt-in to `local`).

src/search/local-provider.ts
```diff
@@ -0,0 +1,164 @@
...
+  private available = false;
+  private embed: EmbedFn | null = null;
+  // key: `${corpus}:${accountId ?? "global"}`
+  private store = new Map<string, StoredItem[]>();
+  private evictionTimer: ReturnType<typeof setInterval> | null = null;
+
+  async initialize(): Promise<void> {
+    try {
+      const { pipeline, env } = await import("@huggingface/transformers");
+      env.cacheDir = "/tmp/hf-cache";
```
skjindal93: **[Blocking-ish] Hardcoded cache dir.** `/tmp/hf-cache` is wiped on restart, so the ~23MB model re-downloads on every cold start (thundering herd across replicas), and breaks entirely if `/tmp` is read-only or there's no egress to HuggingFace. Please make this configurable via an env var with `/tmp/hf-cache` as the fallback, and ideally bake the model into the Docker image so startup needs no network.

src/search/local-provider.ts
```diff
@@ -0,0 +1,164 @@
...
+        const out = await extractor(text, { pooling: "mean", normalize: true });
+        const raw = out.data;
+        return raw instanceof Float32Array ? raw : new Float32Array(raw as ArrayLike<number>);
+      };
+      this.available = true;
+      // Run eviction every 10 minutes
+      this.evictionTimer = setInterval(() => this.evictExpired(), 10 * 60 * 1000);
+      log.info("LocalSearchProvider initialized", { model: EMBEDDING_MODEL, dim: EMBEDDING_DIM });
+    } catch (err) {
+      log.error("LocalSearchProvider initialization failed", { error: String(err) });
```
skjindal93: Init failure is logged at `error` and then fully swallowed, so the server starts "healthy" with search silently dead. Given the provider defaults to `local`, consider surfacing this more loudly (e.g. a startup readiness signal/metric) so a missing dep or failed model load is observable rather than buried in logs.

skjindal93 (commented):
Reviewed the full branch (not just the diff). The design is genuinely solid — clean `SearchProvider` abstraction, good failure isolation (`void ... .catch()` so indexing never breaks callers), null-provider fallback, and the plan doc is honest about its own limitations. Nice work.

Leaving inline comments. Two I'd consider **blocking**, the rest are good follow-ups.

**Blocking:**
1. Feature defaults ON (`HARNESS_SEARCH_PROVIDER=local`) but the engine (`@huggingface/transformers`) is an *optional* dependency + the model downloads at runtime + cache dir is hardcoded to `/tmp`. In a locked-down cluster this either silently no-ops or fails to init on every cold start. Either make the dep required, or default the provider to `none` and opt in.
2. Multi-user tenant-isolation guard is missing. The plan doc itself says `LocalSearchProvider` must not index live customer data (`entities`/`resources` corpus) in multi-user mode, but `harness_list`/`harness_get` fire-and-forget indexing still runs there. Isolation currently works by luck of the wiring; since it's a security property, the guard the plan specifies should land in this PR.

**Non-blocking:** unbounded memory across accounts, redundant re-embedding, routing false-negatives (no safety floor), and missing tests for the routing decision logic. Details inline.

src/tools/harness-list.ts
```diff
@@ -101,6 +102,27 @@ export function registerListTool(server: McpServer, registry: Registry, client:
           }
         }
 
+        // Fire-and-forget: index items for semantic search
+        const provider = searchManager?.getProvider();
+        if (provider?.isAvailable() && isRecord(result) && Array.isArray(result.items)) {
```
skjindal93: **[Blocking] Missing multi-user tenant-isolation guard.** This fire-and-forget indexing into the `resources` corpus runs in multi-user HTTP mode too. The plan doc (`plans/2026-06-26-...md`) explicitly states `LocalSearchProvider` must NOT index live customer data in multi-user mode and even specifies a `canIndexCorpus(corpus, mode)` guard — but it isn't implemented. `initializeIndex` is correctly skipped in multi-user, yet per-call indexing here (and in `harness-get.ts`) is not.

Isolation happens to work today only because each session's `client.account` is baked in via session headers and `setAccountIdResolver` is never used. Since this is a cross-tenant data property, please add the guard (skip `resources` indexing when `mode === multi-user && provider === local`) in this PR rather than deferring it.

src/search/local-provider.ts
```diff
@@ -0,0 +1,164 @@
...
+    if (!this.available || !this.embed) return;
+    try {
+      if (!item.id || !item.content) return;
+      const key = storeKey(item.corpus, item.accountId);
+      if (!this.store.has(key)) this.store.set(key, []);
+      const items = this.store.get(key)!;
+      const existing = items.findIndex(i => i.id === item.id);
+      // Only apply cap for non-permanent items (permanent corpora are bounded by definition)
+      const expiresAt = resolveExpiresAt(item, Date.now());
+      if (existing < 0 && expiresAt !== undefined && items.length >= MAX_ITEMS_PER_KEY) {
```
skjindal93: Memory cap (`MAX_ITEMS_PER_KEY`) only applies to items with a finite TTL — permanent items are exempt, and there's no cap on the *number of keys*. In long-lived multi-user mode you get one `resources:<account>` Map per account (up to ~7.5MB of float embeddings each at cap, plus content strings), unbounded in aggregate as accounts accumulate. Worth a global memory ceiling or cross-key LRU.

src/search/local-provider.ts
```diff
@@ -0,0 +1,164 @@
...
+        // Evict the item with the nearest expiry (soonest to expire = least valuable)
+        let minIdx = 0;
+        for (let i = 1; i < items.length; i++) {
+          const a = items[minIdx]!.expiresAt ?? Infinity;
+          const b = items[i]!.expiresAt ?? Infinity;
+          if (b < a) minIdx = i;
+        }
+        items.splice(minIdx, 1);
+      }
+      const embedding = await this.embed(item.content);
```
skjindal93: Re-embeds on every `index()` call even when the item already exists with identical content. Each call is an ONNX forward pass, and `harness_list` re-indexes its whole result set every time — real CPU for a chatty agent. Cheap win: skip embedding when `existing >= 0 && items[existing].content === item.content`.

src/tools/harness-search.ts
```diff
@@ -75,29 +110,64 @@ export function registerSearchTool(server: McpServer, registry: Registry, client
...
+              log.info(`Semantic search: no high-confidence routing hits, falling back to full scatter-gather`, {
+                query: args.query,
+                results: semanticResults.length,
+                top_score: semanticResults[0]?.score ?? 0,
+              });
+            }
+          }
         }
 
+        const targetTypes = routedTypes ?? candidateTypes;
```
skjindal93, line 180: **Routing false-negatives.** When semantic routing fires, scatter-gather is narrowed to only the predicted types. A confident-but-wrong embedding will skip the correct type, and the user silently gets fewer results than the old full-scan behavior (only signal is `semantic_routed: true`). The no-high-confidence fallback is good, but consider a safety floor: always include tier-1 types (`pipeline`/`service`/`environment`/`connector`) in the routed set, and/or surface `types_skipped` details so the agent can re-query. This is the riskiest behavior change in the PR and has no automated precision test.

src/tools/harness-search.ts
```diff
@@ -44,7 +44,42 @@ interface SearchResultEntry {
...
+/**
+ * Extract unique resource types from semantic results that meet the routing threshold.
+ * Uses both:
+ * - `resources` corpus hits: live Harness data already indexed for this account
+ * - `mcp_resources` corpus hits: resource_definition and entity_schema entries whose
+ *   resource_type field maps directly to a listable API type (e.g. resource-def:connector
+ *   → "connector"). These are the primary routing signal since mcp_resources is always
+ *   indexed at startup whereas resources may be sparse.
+ */
+function extractRoutingTypes(semanticResults: SearchResult[], allTargetTypes: string[]): string[] | null {
```
skjindal93: `extractRoutingTypes` plus the tier-0 semantic merge/dedup are the riskiest new logic and have zero unit tests. They're pure/easily testable with a stub `SearchProvider` (no model needed). Suggest exporting `extractRoutingTypes` and adding tests for: high-confidence single-type routing, multi-type routing, below-threshold → `null` (full fallback), and comma-separated schema types not matching any single type.

src/tools/harness-list.ts
```diff
@@ -101,6 +102,27 @@ export function registerListTool(server: McpServer, registry: Registry, client:
...
+          const accountId = client.account;
+          void Promise.all(
+            (result.items as Array<Record<string, unknown>>).map(item =>
+              provider.index({
+                id: `${resourceType}:${String(item["identifier"] ?? item["id"] ?? "")}`,
+                content: [resourceType.replace(/_/g, " "), item["name"], item["description"], item["identifier"], item["tags"]].filter(Boolean).join(" "),
```
skjindal93: Minor embedding-quality bug: `content` includes `item["tags"]` joined directly, but `tags` from API responses is usually an object/array and stringifies to `[object Object]` in the embedding text. Same pattern in `harness-get.ts` and `manager.ts:initializeIndex`. Consider flattening tags to `key:value` (or just keys) before joining.

thisrohangupta (approved):

PR #570, opened 2026-07-07
fix(http): stop dropping MCP sessions that force re-authentication

thisrohangupta:
## Problem

Remote clients (e.g. the claude.ai Harness connector in Claude Code) are being prompted to re-authenticate very frequently — reported as escalating from "every 5 minutes" to "every prompt."

The connection isn't actually expiring. The server is **losing the session** and returning `404 Session not found` (`src/index.ts:354`), which the MCP client surfaces as a re-authentication prompt. There are two independent, compounding causes.

## Root causes

**1. Multi-replica + per-pod in-memory session store + no session affinity → "every prompt"**

MCP sessions live in a process-local `Map` (`src/index.ts:295`); the SDK transport for a session physically lives in one pod. `k8s/deployment.yaml` ran `replicas: 2` and `k8s/service.yaml` was a plain `ClusterIP` with no `sessionAffinity`. A session initialized on pod A is unknown to pod B, so any follow-up request the LB round-robins to the wrong pod 404s. With 2 replicas this hits roughly half of requests — matching the regression from occasional to near-constant.

**2. Aggressive 5-minute idle TTL → the original "every 5 minutes"**

`MCP_SESSION_TTL_MS` defaulted to `5 * 60_000` and the reaper (`src/index.ts:318`) destroys any session idle beyond it. The connector holds no persistent SSE stream between prompts, so a short pause reaps the session.

## Changes

- **`k8s/deployment.yaml`**: `replicas: 2 → 1`, with a comment explaining the in-memory store constraint and what scaling out requires.
- **`k8s/service.yaml`**: add `sessionAffinity: ClientIP` (timeout matched to the session TTL) as defense-in-depth; note that an L7 ingress must configure its own sticky sessions.
- **`src/config.ts` / `.env.example`**: raise default `MCP_SESSION_TTL_MS` to 30 min (`1800000`).
- **`src/config.ts` + `src/index.ts`**: add `HARNESS_MCP_TRUST_PROXY` and wire `app.set("trust proxy", …)` so per-IP rate limiting keys on the real client behind a load balancer instead of bucketing every user into one counter (`src/index.ts:271`).
- **`AGENTS.md`**: document `HARNESS_MCP_TRUST_PROXY`.
- **`tests/config.test.ts`**: update TTL default assertion; add coverage for `HARNESS_MCP_TRUST_PROXY`.

## Follow-up (not in this PR)

The correct long-term fix for horizontal scaling is externalizing the session/event store (the MCP SDK supports an `eventStore` for resumability). Until then, `replicas: 1` + affinity is the safe configuration.

## Testing

- `pnpm typecheck` — clean
- `pnpm test` — 118 files, 2515 tests passing

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -35,12 +35,18 @@ LOG_LEVEL=info
 # HTTP transport only — ignored in stdio mode
 PORT=3000
 # Idle HTTP sessions are reaped after this many milliseconds once no request
-# or SSE stream is active.
-MCP_SESSION_TTL_MS=300000
+# or SSE stream is active. Default 30m — kept generous so interactive clients
+# (e.g. the claude.ai connector, which does not hold a persistent SSE stream
+# between prompts) are not evicted mid-conversation.
+MCP_SESSION_TTL_MS=1800000
 # Require Authorization: Bearer <token> on /mcp routes when set.
 HARNESS_MCP_AUTH_TOKEN=
 # Non-loopback HTTP binds require HARNESS_MCP_AUTH_TOKEN unless this is true.
 HARNESS_MCP_ALLOW_UNAUTHENTICATED_HTTP=false
+# Number of proxy hops to trust for client IP resolution (Express `trust
+# proxy`). Set to the number of reverse proxies / load balancers in front of
+# the server so per-IP rate limiting keys on the real client. Default 0.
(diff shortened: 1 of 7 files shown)


thisrohangupta:
Per Yogesh this wont solve the issue

PR #676, opened 2026-07-21
fix: register pipeline_error_analysis as alias for debug-pipeline-failure prompt

RajBaviskar:
## Summary
- Registers `pipeline_error_analysis` as an alias for the existing `debug-pipeline-failure` MCP prompt
- Fixes a naming mismatch where ml-infra calls `get_prompt("pipeline_error_analysis")` but the server only registered `debug-pipeline-failure`, causing the prompt lookup to fail silently
- Extracts shared handler/config so both names use identical logic with zero duplication

## Test plan
- [x] `pnpm typecheck` passes
- [x] `pnpm build` succeeds
- [x] All 2607 tests pass (`pnpm test`)
- [x] `pnpm standards:check` passes (80 checks)
- [x] `pnpm docs:check` passes (README up to date)
- [ ] Verify ml-infra can now successfully fetch the prompt via `get_prompt("pipeline_error_analysis")`

### DIFF
--- a/src/prompts/debug-pipeline.ts
+++ b/src/prompts/debug-pipeline.ts
@@ -55,7 +55,11 @@ export function registerDebugPipelinePrompt(server: McpServer): void {
   // Alias: ml-infra requests this name via get_prompt("pipeline_error_analysis")
   server.registerPrompt(
     "pipeline_error_analysis",
-    promptConfig,
+    {
+      ...promptConfig,
+      description:
+        "Alias of debug-pipeline-failure — analyze a failed pipeline execution and suggest fixes. Accepts an execution ID, pipeline ID, or Harness URL.",
+    },
     async (args) => handleDebugPipeline(args),
   );
 }
--- a/src/prompts/debug-pipeline.ts
+++ b/src/prompts/debug-pipeline.ts
@@ -1,29 +1,26 @@
 import * as z from "zod/v4";
 import type { McpServer } from "@modelcontextprotocol/sdk/server/mcp.js";
 
-export function registerDebugPipelinePrompt(server: McpServer): void {
-  server.registerPrompt(
-    "debug-pipeline-failure",
-    {
-      description: "Analyze a failed pipeline execution and suggest fixes. Accepts an execution ID, pipeline ID, or Harness URL.",
-      argsSchema: {
-        executionId: z.string().describe("The failed execution ID, pipeline ID, or a Harness URL").optional(),
-        projectId: z.string().describe("Project identifier").optional(),
-      },
-    },
-    async ({ executionId, projectId }) => {
-      // Detect if the input looks like a URL
-      const isUrl = executionId?.startsWith("http");
-      const idParam = isUrl
-        ? `url="${executionId}"`
-        : `execution_id="${executionId}"`;
+const promptConfig = {
+  description: "Analyze a failed pipeline execution and suggest fixes. Accepts an execution ID, pipeline ID, or Harness URL.",
+  argsSchema: {
+    executionId: z.string().describe("The failed execution ID, pipeline ID, or a Harness URL").optional(),
+    projectId: z.string().describe("Project identifier").optional(),
+  },
+};
 
-      return {
-        messages: [{
-          role: "user" as const,
-          content: {
-            type: "text" as const,
-            text: `Analyze this failed Harness pipeline execution and provide:
+function handleDebugPipeline({ executionId, projectId }: { executionId?: string; projectId?: string }) {
+  const isUrl = executionId?.startsWith("http");
+  const idParam = isUrl
+    ? `url="${executionId}"`
+    : `execution_id="${executionId}"`;
+
+  return {
+    messages: [{
+      role: "user" as const,
+      content: {
+        type: "text" as const,
+        text: `Analyze this failed Harness pipeline execution and provide:
 
 1. **Root cause** of the failure
 2. **Which step failed** and why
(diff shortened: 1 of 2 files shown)


src/prompts/debug-pipeline.ts
```diff
@@ -43,9 +40,22 @@ Then analyze the diagnostic payload:
...
+export function registerDebugPipelinePrompt(server: McpServer): void {
+  server.registerPrompt(
+    "debug-pipeline-failure",
+    promptConfig,
+    async (args) => handleDebugPipeline(args),
+  );
+
+  // Alias: ml-infra requests this name via get_prompt("pipeline_error_analysis")
+  server.registerPrompt(
+    "pipeline_error_analysis",
```
thisrohangupta, line 59: **The alias surfaces a second prompt with an identical description, which is confusing in prompt listings.**

Both `debug-pipeline-failure` and `pipeline_error_analysis` now appear in `listPrompts()` sharing this exact description ("Analyze a failed pipeline execution and suggest fixes…"). The alias only exists so ml-infra can `get_prompt("pipeline_error_analysis")` by name — a direct lookup that never enumerates the list. But any *other* client or agent doing prompt selection now sees two entries that look completely identical, with nothing indicating one is just an alias.

Consider giving the alias a distinguishing description so listing consumers can tell them apart, e.g.:

```
description: "Alias of debug-pipeline-failure — analyze a failed pipeline execution and suggest fixes. Accepts an execution ID, pipeline ID, or Harness URL.",
```

(This means the alias would use its own config object rather than sharing `promptConfig`. Your test at `debug-pipeline.test.ts` asserting `alias.description === original.description` would need updating — but the `getPrompt` *content* equality test, which is the real contract, still holds.)

src/prompts/debug-pipeline.ts
```diff
@@ -1,29 +1,26 @@
...
-        : `execution_id="${executionId}"`;
-
-      return {
-        messages: [{
-          role: "user" as const,
-          content: {
-            type: "text" as const,
-            text: `Analyze this failed Harness pipeline execution and provide:
+const promptConfig = {
+  description: "Analyze a failed pipeline execution and suggest fixes. Accepts an execution ID, pipeline ID, or Harness URL.",
```
thisrohangupta, line 5: **The new alias isn't reflected in the docs, and `docs:check` won't catch it.**

The PR description checks off "`pnpm docs:check` passes (README up to date)", but that check is misleading here: `scripts/generate-docs.js` counts prompt **files** (`readdirSync(src/prompts).filter(f => f.endsWith(".ts") && f !== "index.ts")`), not registered prompts. Since this adds a second registration inside an existing file, the file count stays 33 and the check stays green — but there are now **34 registered prompts**, so the README's "**33 prompt templates**" (README:14) is now inaccurate, and the prompt table (README:1675) still lists only `debug-pipeline-failure`.

Either add a `pipeline_error_analysis` row to the README prompt table (and reconcile the count), or deliberately leave the internal alias undocumented — but if so, note it here so the omission reads as intentional rather than missed.

src/prompts/debug-pipeline.ts
```diff
@@ -43,9 +40,22 @@ Then analyze the diagnostic payload:
...
+}
+
+export function registerDebugPipelinePrompt(server: McpServer): void {
+  server.registerPrompt(
+    "debug-pipeline-failure",
+    promptConfig,
+    async (args) => handleDebugPipeline(args),
+  );
+
+  // Alias: ml-infra requests this name via get_prompt("pipeline_error_analysis")
```
thisrohangupta: **Altitude: `pipeline_error_analysis` (snake_case) diverges from the kebab-case convention every other prompt uses, and this is the second such alias.**

Every prompt in this server is kebab-case (`debug-pipeline-failure`, `create-pipeline`, `onboard-service`, …); `pipeline_summarizer` (PR #648) was the first snake_case exception, and this is the second. Each one is individually justified by "ml-infra calls this exact name," but the pattern is that the server keeps growing aliases to absorb naming mismatches from a caller in another repo.

The comment on line 55 documents *why*, which is good — no change strictly required here. But the durable fix is to align the prompt-name contract across ml-infra and this server (pick one canonical name per prompt) so aliases don't keep accumulating. Worth a tracking item rather than a fix in this PR.

src/prompts/debug-pipeline.ts
RajBaviskar: Addressed — the alias now uses a distinguishing description prefixed with "Alias of debug-pipeline-failure —" (commit 86d2cedd). Test updated to assert the distinguishing description instead of equality.

src/prompts/debug-pipeline.ts
RajBaviskar: Intentionally leaving the alias undocumented in the README prompt table — it's an internal lookup alias, not a distinct template (same precedent as `summarize-pipeline` alias for `pipeline_summarizer`). Added an inline comment making this deliberate (commit ecf3bc9e). The "34 prompt templates" count is file-based so it remains accurate.

src/prompts/debug-pipeline.ts
RajBaviskar: Agreed — no naming change in this PR. Tracking issue #677 covers aligning ml-infra and mcp-server on one canonical name per prompt. Added the issue link in the inline comment so it's visible at the alias site.

src/prompts/debug-pipeline.ts
RajBaviskar: Addressed — the alias now uses a distinguishing description prefixed with "Alias of debug-pipeline-failure —" (commit 86d2cedd).

thisrohangupta (approved):

PR #800, opened 2026-08-12
feat: add IaCM variable set list, get, create, and update

anushk-singhal:
## Summary
- Add multi-scope `iacm_variable_set` (account/org/project) with list/get/create/update against existing IaCM APIs
- Create/update return the VariableSet resource (not a `policy_evaluation` envelope); writes are `medium_write` with confirmation
- Document that `iac_variableset_*` RBAC permissions are still **Experimental** in Harness — deny paths are not enforceable until iac-server activates them; MCP continues to forward the caller PAT/SAT unchanged
- Stacked on merged workspace support (#793)

## Test plan
- [x] Unit: variable-set contract, list extract, create/update validation, multi-scope pathBuilder dispatch
- [x] Mock-fetch integration for create/update URL + body
- [x] MCP `harness_create` / `harness_update` + medium_write elicitation
- [x] README resource table + Experimental RBAC note
- [ ] Optional live smoke with SATs (expect allow on view/none/edit while Experimental)

### DIFF
--- a/README.md
+++ b/README.md
@@ -1422,13 +1422,16 @@ Template operations use the Harness Template service paths (`/template/api/templ
 
 ### Infrastructure as Code Management (IaCM)
 
-IaCM resources are default-enabled and mostly project-scoped. Start with `iacm_workspace` to find workspace identifiers, then use that `workspace_id` for workspace resources, costs, and activity diffs. The module registry is account-scoped.
+IaCM resources are default-enabled and mostly project-scoped. Start with `iacm_workspace` to find workspace identifiers, then use that `workspace_id` for workspace resources, costs, and activity diffs. Use `iacm_variable_set` for reusable variable sets at account, org, or project scope. The module registry is account-scoped.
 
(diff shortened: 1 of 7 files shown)


anushk-singhal:
Addressed the update-semantics ask (and the two easy follow-ups):

**Update semantics (confirmed against iac-server)**  
`UpdateVariableSet*` uses `mergeVariableSetVars` / connectors / files. Omitting a collection (or sending `{}` / `[]`) **clears** prior entries — it is not PATCH-preserve. So the old “replace when provided” wording was wrong and unsafe for agents.

**Fix in this push**
- Document HTTP PUT + full-replacement + **get-then-put** on the resource, update op, bodySchema, and README
- Require `terraform_variables` and `environment_variables` on update (send maps from `harness_get`, or `{}` to clear) so agents cannot silently wipe vars by sending `{ name }` only
- Nest `key` / `value` / `value_type` field specs on those maps (reuse workspace pattern)
- Add account-scope create/get path test; update tests no longer send name-only bodies

**Deferred (as you noted, non-blocking)**  
DELETE at all three scopes — intentional for this stacked PR.

yogesh-chauhan (approved):

PR #811, opened 2026-08-13
feat: add IaCM provider registry create and version update

anushk-singhal:
## Summary
- Adds `iacm_provider` list/get/create and version-oriented update via the declarative registry.
- **Create:** `POST /iacm/api/providers/{type}` — `body.type` is the path segment; optional `description` in JSON. Response is `{ id }` only — follow up with `harness_get` / `harness_list`.
- **Update:** version-only (no metadata PUT) — POST `/providers/{id}/version` to create a version, or PUT `/providers/{id}/version/{version}` when `params.version` is set. Required: `protocol` + `gpg_key_id` (+ `body.version` when creating). Empty 201/204 bodies normalize to `{ status: "SUCCESS", message: "No content" }` via HarnessClient.
- **Scope:** account-only by IaCM API design (no `scope_org`/`scope_project`, unlike `iacm_module`). Ambient `org_id`/`project_id` are ignored; `resource_scope=org|project` is rejected.
- Documents Experimental provider-registry RBAC (`iac_providerregistry_*`).

## Coverage
- Registry contract, mock-fetch (create `{ id }`, empty version bodies, ambient org/project leak regression), tool handlers, elicitation, HarnessClient empty-2xx.

## Test plan
- [x] `pnpm typecheck`
- [x] `pnpm test` (130 files, 2868 passing)
- [x] `pnpm build && pnpm docs:generate`
- [ ] Optional QA smoke with Experimental RBAC note (deny paths not enforceable)


anushk-singhal:
Thanks for the review — all three notes are non-blocking:

1. Empty 2xx → success is intentional for IaCM version writes (201 with no body). We’ll keep the current behavior; 200/204 coverage + a debug log can be a follow-up.

2. Elicitation mock should be{ id }only — same contract as the other provider tests. We’ll fix in a follow-up if we don’t squeeze it in here.

3. Version-missing preflight test is a good sibling to modulename/system; tracked as follow-up.

thisrohangupta (approved):

PR #873, opened 2026-08-28
add HAR v3 read-only toolset

Sourabh-Awasthi:
## Summary

Adds a new `registries-v3` toolset covering the Artifact Registry v3 standardized API (reads only): packages, versions, files, per-entity metadata, metadata key/value discovery, scans (list + details), bulk scan evaluation status, and firewall exceptions.

v3 uses flat paths with **snake_case scope query params** (`account_identifier` / `org_identifier` / `project_identifier`), so a dedicated `scopeParams` override is threaded through each resource — this is the main structural difference from v1's path-based `/har/api/v1/spaces/{acct}/{org}/{proj}/+/…` refs.

Adds `harV3ListExtract` which normalizes both response shapes v3 endpoints actually return today:
- Native v3: `{ items, page, size, hasMore, meta }`
- Legacy v1 envelope (still returned by `/scans`): `{ data, itemCount, pageIndex, pageSize, pageCount }`

Verified against QA (`qa.harness.io/gateway/har/api/v3`) — packages, versions, files, metadata/keys, metadata/values, scans, and scans/exceptions all return the expected shape via the toolset. A regression test covers the `/scans` v1-envelope normalization.

**Scope: reads only.** Writes (firewall exception create/update/status, metadata upsert/save, tag add, bulk-evaluate, copy) and internal ops (delete/restore/migrate, backfill, file preview/search, bulk download) are intentionally deferred to follow-up PRs.

Tracks [AH-5245](https://harness.atlassian.net/browse/AH-5245).

## Test plan

- [x] `pnpm test tests/registry/` — 480/480 pass (479 pre-existing + 1 new normalization test)
- [x] `pnpm typecheck` — clean apart from the pre-existing unrelated `@resvg/resvg-js` error on `main`
- [x] Manual smoke against QA `/har/api/v3` — verified live 200s for `packages`, `versions`, `files`, `metadata/keys`, `metadata/values`, `scans`, `scans/exceptions`, `scans/versions`
- [ ] Follow-up PR 2 will add v3 writes + stats endpoints

[AH-5245]: https://harness.atlassian.net/browse/AH-5245?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/README.md
+++ b/README.md
@@ -502,13 +502,10 @@ pnpm docker:run
 docker run --rm -p 3000:3000 \
   -e HARNESS_API_KEY=pat.xxx.xxx.xxx \
   -e HARNESS_ACCOUNT_ID=your-account-id \
-  -e HARNESS_MCP_AUTH_TOKEN=replace-with-a-long-random-token \
   harness-mcp-server
 ```
 
 The container runs in HTTP mode on port 3000 by default with a built-in health check.
-It binds to the container network, so set `HARNESS_MCP_AUTH_TOKEN` in `.env` or the
-container environment and send that value as `Authorization: Bearer <token>` on MCP requests.
 
 ### Kubernetes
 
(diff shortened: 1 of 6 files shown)


yogesh-chauhan (approved):

yogesh-chauhan (changes requested):
Please take a look at Comments from Cursor Architecture Review

yogesh-chauhan (approved):

PR #892, opened 2026-09-03
refactor: rename module-coded resources & toolsets to product capabilities (non-breaking)

thisrohangupta:
## What & why

Harness resource types and toolset filter tokens were coded to internal module/service names and acronyms (`iacm`, `sto`/`scs`, `ccm`, `fme`, `sei`, `idp`, `dbops`, `knowledge-graph`/`semantic-layer`). Agents and users pick tools by product capability, not by which backend service happens to own an endpoint. This renames every such identifier to the product capability it represents, and standardizes the four remaining hyphenated toolset tokens to snake_case so the entire public surface (tool names, resource types, toolset filters) is consistent.

**All of it is non-breaking** — a declarative alias layer keeps every old `HARNESS_TOOLSETS` value and every old `type` param resolving to the renamed canonical entity, with load-time collision guards.

## Renames

**Toolsets → capability**
- `iacm` → `infrastructure` (CD infra-def folded into `environments` as `infrastructure_definition`)
- `sto` + `scs` → `application_security` (merged)
- `ccm` → `cost`, `dbops` → `databases`
- `sei` → `developer_insights`, `idp` → `developer_portal`
- `fme` → `feature_flags`; `knowledge-graph` + `semantic-layer` → `software_delivery_knowledge_graph` (merged)

**Toolset tokens → snake_case** (hyphen names kept as aliases)
- `pull-requests` → `pull_requests`, `evidence-vault` → `evidence_vault`, `ai-evals` → `ai_evals`, `release-management` → `release_management`

Resource types were canonicalized to match (e.g. `fme_feature_flag` → `feature_flag`, `iacm_variable_set` → `infrastructure_variable_set`), and internal string-comparison sites that the alias layer doesn't cover (URL parser, search priority weights) were updated to the canonical names.

## Convention

Underscore, to match the surface that was already snake_case (`harness_list`, `feature_flag`). GitHub's MCP server and Anthropic's reference servers use snake_case; hyphens weren't anyone's tool-name convention.

## Docs

Regenerated the README from the registry: 38 toolsets, canonical resource types, and the full Toolset Filtering table (which had drifted stale across the whole rename). Re-enabled coverage validation for `release_management` and `feature_flags` (both had silently fallen out of the validated set when their curated keys went stale), so those matrices can't drift unnoticed again. `.env.example` and `AGENTS.md` updated with canonical names + a legacy-alias note.

## Verification

`pnpm test` (3259 passing), `pnpm standards:check` (77 passing), `pnpm docs:check` clean. Added `it.each` alias regression tests covering every old hyphen/module name → canonical resolution.


sunilgattupalle:
## Review

Pulled the branch and verified independently rather than trusting the PR description: `pnpm typecheck` clean, `pnpm test` → 3319/3319 passing, `pnpm standards:check` → 77/77, `pnpm build && pnpm docs:check` → README up to date, registry loads cleanly (243 resource types / 37 toolsets).

**Implementation**
- Alias mechanism is solid: `resolveType()` applied at every lookup choke point (`getResource`, `supportsOperation`, `getExecuteActions`, `auditBlockedAttempt`), never enumerated by `getAllResourceTypes()`, with load-time guards for alias-shadows-canonical and alias-claimed-twice.
- Went beyond the mechanical rename — `searchAliases` (e.g. `"sto"`) were added per-resource so the old module code still surfaces in `harness_search` even after the canonical name drops it.
- Spot-checked the `application_security` merge (all 27 former STO+SCS resources): cross-reference text inside `relatedResources` graphs was updated too, not just top-level `resourceType` keys.

**Naming quality** — glad to see `fme` landed on `feature_flags` (matches the toolset's actual `displayName`, "Feature Management & Experimentation") rather than an invented name, and that `chaos`/`resilience` wasn't touched here — "chaos engineering" is industry-standard vocabulary, not an internal module code, so renaming it away would likely reduce agent recognition rather than improve it. Worth keeping that lens for any future toolset renames: the test isn't "is this a short/clean identifier" but "does this match displayName / is this already a term agents already know."

**One gap, now filled in a stacked PR**: the alias layer guards alias-vs-canonical and alias-vs-alias collisions, but nothing guarded against two toolsets declaring the same *canonical* `resourceType` — `resourceMap.set()` would silently overwrite one with no error. Given the scale of manual renaming here (~65 identifiers, two large files merged into one), that's exactly the kind of typo that's easy to introduce and easy to miss in review. Opened #896 (stacked on this branch) adding `assertUniqueResourceTypes()` at Registry construction + 3 regression tests — happy to fold it in directly if you'd rather not merge it as a separate PR.

PR #983, opened 2026-09-19
Prototype: dynamic blast-radius risk scoring (spec 007)

sunilgattupalle:
## Summary
- Prototypes spec 007 for a single pilot: `pipeline.delete` can now have its effective risk lowered from `destructive` toward `high_write` when a TypeSafe Score call confidently rates the specific pipeline as low blast-radius (never run / stale, no prod tag).
- Adds a minimal TypeSafe HTTP client (`src/client/typesafe-client.ts`) for the `Score` primitive only — there is no TypeScript `typesafe-sdk` yet, so the wire contract (`POST /v1/systemone`) was reverse-engineered from the Python SDK's source in `harness-evals`.
- `scoreEffectiveRisk()` (`src/utils/risk-scoring.ts`) is fail-closed: disabled, unregistered, timeout, error, or low-confidence all fall back to the static risk — identical to today's behavior. The feature is entirely opt-in via `HARNESS_DYNAMIC_RISK_SCORING` (default `false`) and requires `TYPESAFE_API_KEY`.
- Audit rows and the tool response gain a `risk_scoring`/`risk_assessment` field only for calls where the scorer actually ran.
- Verified against a real TypeSafe key: 3 hand-built pipeline states (never-run, stale/no-prod, active/prod-tagged) scored correctly with confidence 1.0, latency 148–391ms — comfortably inside the 400ms default timeout.

## Scope (deliberately narrow — this is a prototype, not the full spec)
- Only `pipeline.delete` is wired. `feature_flag.update` (the spec's second pilot) is not done.
- No integration tests yet, only unit tests for `scoreEffectiveRisk()` (9 tests, including the fail-closed property test).
- Prod-environment detection in the scorer is a crude heuristic (`JSON.stringify(...).includes("prod")`) — flagged in code as illustrative/to-be-refined, matching the spec's own caveat.
- Only 3 manually-constructed states have been tried against the real API — no load test, no adversarial/near-threshold cases.

## Test plan
- [x] `pnpm typecheck && pnpm build && pnpm test` (3717/3718 passing; the one failure is a pre-existing, unrelated symlink-extraction test)
- [x] Live smoke test against a real `TYPESAFE_API_KEY` (see summary above)
- [ ] `feature_flag.update` pilot
- [ ] Integration tests per the spec's test plan
- [ ] p95 latency under real concurrency


sunilgattupalle:
Superseded — consolidated into #984, which is now retargeted at `main` and contains both pilots (this PR's commits plus the repo_rule pilot). Closing per feedback that a single combined PR is preferable to a stack.

PR #984, opened 2026-09-19
Prototype: dynamic blast-radius risk scoring for pipeline.delete and repo_rule (specs 007-008)

sunilgattupalle:
## Summary

Two pilots for spec 007's dynamic blast-radius risk scoring mechanism, combined into one PR (previously stacked as separate PRs #983/#984 — consolidated per review feedback).

- **Mechanism (spec 007):** `OperationPolicy` gains an optional `riskFloor` — a lower-than-`risk` bound. A registered `RiskScorer` can pull a specific call's effective risk down toward that floor using a live TypeSafe `Score` call; it can never raise risk above the static `risk` ceiling. `scoreEffectiveRisk()` (`src/utils/risk-scoring.ts`) is fail-closed: disabled, unregistered, timeout, error, or low-confidence all resolve to `effectiveRisk = risk` — identical to today's unchanged behavior. Entirely opt-in via `HARNESS_DYNAMIC_RISK_SCORING` (default `false`) and requires `TYPESAFE_API_KEY`. Adds a minimal TypeSafe HTTP client (`src/client/typesafe-client.ts`) — no TypeScript `typesafe-sdk` exists yet, so the wire contract (`POST /v1/systemone`) was reverse-engineered from the Python SDK in `harness-evals`.
- **Pilot 1 — `pipeline.delete`:** effective risk lowered from `destructive` toward `high_write` when the scorer confidently rates the specific pipeline as low blast-radius (never run / stale, no prod tag). **Verified against a real TypeSafe key**: 3 hand-built states (never-run, stale/no-prod, active/prod-tagged) scored correctly, confidence 1.0, latency 148–391ms.
- **Pilot 2 — `repo_rule.update`/`repo_rule.delete`:** same one-level-floor pattern (`high_write`→`medium_write`, `destructive`→`high_write`), sharing one scorer that grades blast radius from the rule's enforcement state (active/disabled/monitor), whether it targets the default branch, and recent commit activity (last 30 days, capped read). Also fixes a real gap found while building this: `harness_update` was never wired into `scoreEffectiveRisk` at all (only `harness_delete` was), so any `riskFloor`/`riskScorer` on an `update` op was silently inert.
- **`specs/008-decision-primitives-inventory.md`:** the survey behind pilot 2 — where Score/Noul/Choice primitives are realistically applicable across the server, explicit anti-force-fit criteria, a normative fallback contract, and the concrete implementation spec (including a "review findings" section documenting bugs 1–4 below and their fixes).

## Review findings (independent Opus review of pilot 2, all fixed before this PR)
1. **(High)** Both raw `client.request()` GETs in the `repo_rule` scorer omitted org/project scope params → 404 on any project-scoped repo, the common case, not an edge case. Fixed: pass `orgIdentifier`/`projectIdentifier` explicitly.
2. **(High)** `since` sent in milliseconds; the Code API's commits endpoint expects unix **seconds** → read as a timestamp tens of thousands of years out → empty commit window → biases blast radius *downward* for busy repos, the wrong direction for a safety mechanism. Fixed: converted to seconds.
3. **(Medium)** Two sequential GETs plus the TypeSafe call didn't fit pilot 1's single-GET 400ms timeout budget. Fixed: parallelized the two GETs via `Promise.all`.
4. **(Medium)** `ctx.signal` was never populated by either caller, so a scorer timeout didn't actually cancel in-flight requests. Fixed generically in `scoreEffectiveRisk()` itself (benefits both pilots): constructs its own `AbortController`, aborts on the timeout/error path.
5. **(Medium)** The scorer rated the rule's state *before* the requested change — re-enabling a disabled rule on the default branch (exactly the high-blast-radius case this exists to catch) would have scored the old, disabled state. Fixed: the requested PATCH body's `state`/`pattern` now override the fetched values when present.

Two points raised by the review were deliberately **not** changed: reusing pilot 1's config flag rather than adding a new one (accepted as a documented deviation — see spec 008, same mechanism not a new site shape), and no new tool-level regression test for `harness-update.ts`'s wiring (neither `harness-update.ts` nor `harness-delete.ts` has any existing tool-level test coverage, so adding one only now would be inconsistent scope expansion).

## Scope (deliberately narrow — this is a prototype, not the full spec)
- Only `pipeline.delete` and `repo_rule.update`/`.delete` are wired. `feature_flag.update` (spec 007's originally-listed second pilot) turned out to already sit at `risk: "low_write"` in the current codebase — below the floor the spec assumed — so it's flagged in spec 008 as a spec/implementation mismatch requiring a separate risk-posture decision, not built here.
- No integration tests; only unit tests for the generic `scoreEffectiveRisk()` engine (9 tests + a fail-closed property test), matching the existing test-coverage bar for `harness-delete.ts`/`harness-update.ts` (neither has tool-level tests today).
- Prod-environment detection in the `pipeline.delete` scorer is a crude heuristic (`JSON.stringify(...).includes("prod")`) — flagged in code as illustrative/to-be-refined.
- **Pilot 2 has zero live TypeSafe validation** — no API key was available while building it. It's typecheck/build/unit-test verified only; pilot 1's live results should not be read as evidence for pilot 2's accuracy or latency.
- The confidently-wrong-scorer gap is named, not mitigated: fail-closed protects against an *unavailable* score, not a *confidently wrong* one.

## Test plan
- [x] `pnpm typecheck && pnpm build && pnpm test` — 3717/3718 passing; the one failure (`adm-zip-security-lib.test.ts` symlink test) is pre-existing and unrelated (confirmed via `git status` — file untouched)
- [x] `pnpm standards:check` — 79/79 passing
- [x] Live smoke test against a real `TYPESAFE_API_KEY` for pilot 1 (3/3 correct, 148–391ms)
- [ ] Live smoke test for pilot 2 (`repo_rule`) — not yet run
- [ ] `feature_flag.update` pilot — blocked on a separate risk-posture decision (see spec 008)
- [ ] Integration tests per spec's test plan
- [ ] p95 latency under real concurrency


sunilgattupalle:
An independent review flagged bugs serious enough that the pilot would have been a silent no-op in practice. Fixed in the latest commit:

1. **Missing org/project scope on both scorer GETs** — they bypass the registry's automatic scope injection (raw `client.request()` calls), so a project-scoped repo 404'd, fell back to static risk, and the pilot never actually scored anything on the common path. Fixed: pass `orgIdentifier`/`projectIdentifier` explicitly, matching pilot 1's pattern.
2. **`since` unit mismatch** — passed milliseconds where the Code API expects unix seconds, reading as a timestamp far in the future → empty commit window → biased blast radius *downward* for busy repos, the wrong direction for a safety mechanism. Fixed.
3. **Timeout budget didn't fit** — two sequential GETs + the TypeSafe call inside the same 400ms budget pilot 1 uses for one GET + one TypeSafe call. Fixed by running the two GETs concurrently (`Promise.all`).
4. **Correctness bug**: the scorer rated the rule's *current* state, not the requested change — re-enabling a disabled rule on the default branch (the exact high-blast-radius case this exists to catch) would have scored the old, disabled state. Fixed: PATCH body fields override the fetched state when present.
5. Also: `scoreEffectiveRisk` now aborts its own `AbortController` on timeout/error, so a scorer's in-flight requests don't keep running after the caller has already fallen back to static risk.

`specs/008-decision-primitives-inventory.md`'s pilot 2 section now documents all of this. `pnpm typecheck`/`build`/`test`/`standards:check` all still green (same one pre-existing unrelated failure).

